PR #30, opened 2013-08-16
added nginx-convention for IP-checking

a7p:
Nginx by default forwards the IP as "HTTP_X_REAL_IP" - I added checking for this.

### DIFF
--- a/analytical/utils.py
+++ b/analytical/utils.py
@@ -115,7 +115,9 @@ def is_internal_ip(context, prefix=None):
     """
     try:
         request = context['request']
-        remote_ip = request.META.get('HTTP_X_FORWARDED_FOR', '')
+        remote_ip = request.META.get('HTTP_X_REAL_IP', '')
+        if not remote_ip:
+            remote_ip = request.META.get('HTTP_X_FORWARDED_FOR', '')
         if not remote_ip:
             remote_ip = request.META.get('REMOTE_ADDR', '')
         if not remote_ip:


hobarrera:
`HTTP_X_FORWARDED_FOR` may be a list of IPs, comma separated (eg: ` 'HTTP_X_FORWARDED_FOR': '167.114.234.31, 10.0.10.90',`).  We need to keep the first one.

jcassee:
Well, unless you have no proxy yourself. Then even the first one is spoofed. I think it must be configurable at least.

jcassee:
@a7p @max-arnold @hobarrera Any of you interested in pursuing this pull request?

hobarrera:
Sorry, not me. I've currently no need for this, and my free time is nonexistant. :(

a7p:
What needs to be done to close this?

* introduce a setting to activate caring about IP-HEADERS ()
* introduce a setting to set the ip-headers name
-> or just: USE_IP_FROM_HEADER = "HTTP_X_REAL_IP"

+ Tests for those settings

would that be appropiate to consider this fixed?

jcassee:
I like the way Nginx handles these headers: http://nginx.org/en/docs/http/ngx_http_realip_module.html

Basically, they make you set the IP address of your proxy (or addresses of your proxies) and only honor the headers from those addresses.

EvaSDK:
You might want to check how django-axes implemented it. It looks like it is a sane base.

jcassee:
@EvaSDK Thanks for the link. Are you interested in contributing this functionality?

vdboor:
I'm not such a big fan of *django-ipware* (that django-axes uses too). In it's default configuration, it checks all possible headers, allowing the IP to be spoofed easily. Using something like [wsgi-unproxy](https://bitbucket.org/kwi/wsgiunproxy/) protects against this. As extra benefit, it makes sure `request.META['REMOTE_ADDR']` contains the actual IP address so all packages (including sentry/raven) start reporting the proper IP address.

EvaSDK:
The use of django-ipware is kinda new in django-axes so I cannot comment on what it does good or bad yet. However, Imho, your reverse proxy, should do the work of wiping headers that are used by your application to detect the IP address of the client. It is pretty easy to do and avoids having to figure out how to pipe additional software or configuration to the application.

EvaSDK:
For the record, I just audited django-ipware when trying to upgrade django-axes in my company's codebase and it is currently unsuitable for consumption behind reverse proxies. See https://github.com/jazzband/django-axes/issues/286

PR #48, opened 2014-08-26
Identifying authenticated users in Uservoice

mgaitan:
This PR adds support to indentigy users in the Uservoice service. it Includes some docs


jcassee:
I released version 0.20.0 with your fixes. Sorry it took so long!

PR #49, opened 2014-09-05
Add Django 1.7 support

craigbruce:
* Added Django 1.7 to tox/travis (confirmed tox works, guessing for travis).
* Test failures became apparent in clickmap which have been fixed.
* Removed override_settings to django function instead (also removed with/without_apps as rarely used), required for Django 1.7 as the code copied from django does not honor the new app loading system.


jcassee:
I released version 0.20.0 with your changes. Sorry it took so long!

PR #52, opened 2015-04-18
Adding python 3 compatibility

amadornimbis:
Fixed the exception problem that was presenting problems for python >= 3.0. Changed some testing assertions, where possible, that have been deprecated in new versions of python. All of the tox tests are passing, as are all builds in the Travis build matrix.


jcassee:
Thanks, Eric! I will do a release shortly.

Issue #54, opened 2015-04-20
Test with Django 1.8


jcassee:

bittner:
Can't we just add another line to the test matrix in the [Travis file](/jcassee/django-analytical/blob/master/.travis.yml), and trigger the test suite?

jcassee:
Indeed, that is all, but I needed a place to empty my brain.

jcassee:
Fixed by #55

PR #55, opened 2015-04-20
Also test with Django 1.8 (Travis-CI)

bittner:
Plus white-space changes to match typically used YAML syntax (indentation of lists).

This commit may close issue #54 if all goes well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,26 @@
 language: python
 python:
-- "2.6"
-- "2.7"
-- "3.3"
-- "3.4"
+  - "2.6"
+  - "2.7"
+  - "3.3"
+  - "3.4"
 env:
-- DJANGO="Django>=1.4,<1.5"
-- DJANGO="Django>=1.5,<1.6"
-- DJANGO="Django>=1.6,<1.7"
-- DJANGO="Django>=1.7,<1.8"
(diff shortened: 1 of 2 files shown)


jcassee:
Cool! Could you also update `tox.ini`? It's a bit of a nuisance having to keep them in sync, but I use it locally for testing.

bittner:
I've refactored `tox.ini` and `.travis.yml` based on the following information:

- [Generating environments, conditional settings](http://tox.readthedocs.org/en/latest/config.html#generating-environments-conditional-settings), Tox 1.9 documentation
- [Using Tox with Travis CI to Test Django Apps](http://jsatt.com/blog/using-tox-with-travis-ci-to-test-django-apps/) by Jeremy Satterfield

I didn't find information about Travis-CI being able to handle Bash-like variable expansion (conditional settings), so the list of test environments in `.travis.yml` needs to be generated using:
``` bash
$ tox -l | xargs -I ITEM echo "  - TOXENV="ITEM
```

jcassee:
Looks great! I don't know whether Travis supports expansion, but this is already pretty nice. Thanks!

jcassee:
Ah, we forgot to update the changelog. I'll update it in a direct commit to master, alright?

PR #57, opened 2015-04-25
Fix flake8 complaints (except for templatetags and tests)

bittner:
These changes fix mainly whitespace complaints by `flake8` everywhere except in the `templatetags` und the `tests` folder, i.e.
```
$ flake8 --exclude=templatetags,tests
```
runs without complaints.


jcassee:
I don't care much for PEP8's indenting rule concerning line continuations, but a standard is a standard. :-)

Issue #59, opened 2015-05-15
Not compatible with python 3?


mgaitan:
I was trying to setup analytics for [Walik](https://github.com/mgaitan/waliki)'s demo and it doesn't work with python 3.4. 

```
(waliki)tin@morochita:~/lab/waliki$ pip install django-analytics
Downloading/unpacking django-analytics
  Downloading django-analytics-0.0.1.tar.gz
  Storing download in cache at /home/tin/.pip_download_cache/https%3A%2F%2Fpypi.python.org%2Fpackages%2Fsource%2Fd%2Fdjango-analytics%2Fdjango-analytics-0.0.1.tar.gz
  Running setup.py (path:/home/tin/.virtualenvs/waliki/build/django-analytics/setup.py) egg_info for package django-analytics
    
Downloading/unpacking django-geckoboard>=1.1.0 (from django-analytics)
  Downloading django-geckoboard-1.2.8.tar.gz
  Storing download in cache at /home/tin/.pip_download_cache/https%3A%2F%2Fpypi.python.org%2Fpackages%2Fsource%2Fd%2Fdjango-geckoboard%2Fdjango-geckoboard-1.2.8.tar.gz
  Running setup.py (path:/home/tin/.virtualenvs/waliki/build/django-geckoboard/setup.py) egg_info for package django-geckoboard
    
Installing collected packages: django-analytics, django-geckoboard
  Running setup.py install for django-analytics
    
      File "/home/tin/.virtualenvs/waliki/lib/python3.4/site-packages/analytics/maintenance.py", line 60
        print _("Scanning %(app)s for metrics...") % {'app': import_app}
              ^
    SyntaxError: invalid syntax
    
      File "/home/tin/.virtualenvs/waliki/lib/python3.4/site-packages/analytics/management/commands/metrics.py", line 103
        print _("Metric(s) %(metric)s activated.") % {'metric': kwargs['activate']}
              ^
    SyntaxError: invalid syntax
    
      File "/home/tin/.virtualenvs/waliki/lib/python3.4/site-packages/analytics/geckoboard_views.py", line 25
        raise Exception, _("No array called '%(varname)s' in GET variables") % {'varname': var_name}
                       ^
    SyntaxError: invalid syntax
    
      File "/home/tin/.virtualenvs/waliki/lib/python3.4/site-packages/analytics/calculator.py", line 40
        print _("No data for metric %(metric)s, skipping.") % {'metric': metric.title}
              ^
[...]

PR #60, opened 2015-05-15
mark the package python 3 compatible

mgaitan:
Part of my confusion in #59 was to see no reference to python 3 in `setup.py` . Added

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,8 @@ setup(
         'Topic :: Software Development :: Libraries :: Python Modules',
         'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.3',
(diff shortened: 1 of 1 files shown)


bittner:
Thanks for the changes. Yes, according to the tests django-analytical [works with Python 3](https://github.com/jcassee/django-analytical/blob/master/tox.ini#L12). Good that from now on this is also reflected on PyPI.

bittner:
@jcassee should we upload a new, minor, release? Would you?

[The changes](https://github.com/jcassee/django-analytical/compare/v0.21.0...HEAD) are not too relevant, only one very minor code change (clickmap). It would mostly be to update the documentation and meta-information on PyPI. The docs could also be regenerated.

jcassee:
I release version 0.22.0.

PR #61, opened 2015-05-18
Calculate coverage and send to Coveralls.io

jcassee:
We have pretty great test coverage, so let's advertise that!

This change does drop the `-Wall` option to `python setup.py test`. I could not determine how to tell the `coverage` script how to pass it to `python`. Is that important?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+source = analytical
+omit = analytical/tests/*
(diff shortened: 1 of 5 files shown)


jcassee:
I have (hopefully) disabled the Coveralls comments.

bittner:
Just for sake of documentation:

> It’s a good idea to run your tests with Python warnings enabled: `python -Wall manage.py test`. The `-Wall` flag tells Python to display deprecation warnings.

(from the [Testing section](https://docs.djangoproject.com/en/1.8/topics/testing/overview/) of the Django documentation)

PR #66, opened 2015-09-13
Update locations of constants in Google Analytics documentation

ericdwang:
Looks like the `google_analytics` module was moved into `templatetags` some time ago and the documentation wasn't updated with it.


jcassee:
@bittner: Well, if the correct location is `analytical.templatetags.google_analytics` then the documentation should be updated, right?

PR #70, opened 2015-09-25
Also test Python 3.2 with tox

bittner:
Match python/django test combinations with Pip packages on PyPI:

1. Add Python 3.2
1. Remove Python 3.4 for non-supported Django versions

## Reference

* [Django 1.4](https://pypi.python.org/pypi/Django/1.4#downloads) package (PyPI)
* [Django 1.5](https://pypi.python.org/pypi/Django/1.5#downloads) package (PyPI)
* [Django 1.6](https://pypi.python.org/pypi/Django/1.6#downloads) package (PyPI)
* [Django 1.7](https://pypi.python.org/pypi/Django/1.7#downloads) package (PyPI)
* [Django 1.8](https://pypi.python.org/pypi/Django/1.8#downloads) package (PyPI)


bittner:
:newspaper: **Bad News:** [coverage](https://pypi.python.org/pypi/coverage#last) (:exclamation:) doesn't support Python 3.2! We'd have to run the tests without `coverage` to make them pass. Nasty.

@jcassee What do you think should we do?

jcassee:
Looking at [the last comment in the bug report](https://bitbucket.org/ned/coveragepy/issues/407/coverage-failing-on-python-325-using#comment-21995756) it may get fixed.

Anyway, there is little (not none, though) reason for running coverage for all Python versions. We can opt to run coverage only for the last Python 2 and 3 versions.

bittner:
Pinning the coverage package we're now getting real testing failures with Python 3.2 (no `u`-prefix in strings allowed in Python 3.2). I'll try to fix that.

PR #71, opened 2015-11-05
Fix link to Gaug.es in README

JamesPaden:
Gauges homepage requires the get. subdomain.


jcassee:
Thanks for the fix, James!

Issue #72, opened 2015-11-20
New release?


hobarrera:
It's been a while since the last release, and the latest `master` crashes with django19 (which is at rc, and will be release in less than a couple of weeks).

Can we get a new release (1.0.1?), that's django19-compatible? :)

bittner:
All our builds are passing. As long as Django 1.9 is not stable don't expect it to be added to the build matrix.

What is crashing? Can you post a stack trace? Can you provide a PR?

hobarrera:
Django 1.9 has already been stabilized, and is currently on rc1.

`master` works with django1.9, but the latest release isn't working, that why I was merely mentioning a new release and didn't give much more details. :D  
Notably, 87bb09c fixes a deprecation from 1.7 that no longer works on 1.9, but that didn't make it into 1.0.0.

bittner:
I'd recommend you add 1.9 to the [build matrix](https://github.com/jcassee/django-analytical/blob/master/tox.ini), and we see what happens. If all goes well with the code as it is. Your PR is appreciated.

As soon as 1.9 is officially out we'll merge that into `master` and package a release, I'd suppose.

hobarrera:
I've added 1.9 to the dependency matrix (along with py35 which was also missing, and has been release for some time now).

Hopefully this can be merged in and a new release can be made? :D

brainrape:
bump

bittner:
@brainrape See [our Gitter chat](https://gitter.im/jcassee/django-analytical). Joost must take a decision on the version number. We're getting there.

brainrape:
Thanks

jcassee:
Finally release version 2.0.0 that is compatible with 1.9 thanks to @hobarrera.

PR #73, opened 2015-11-20
Django 1.9 support (dropping unsupported versions)

hobarrera:
See #72


hobarrera:
All tests have passed, but coveralls seems to not be entirely configured:

    You have to provide either repo_token in .coveralls.yml, or launch via Travis or CircleCI

I'm not very familiar with the propietary platform, so haven't looked into it, though I *have* added some code so that tests don't fail if coveralls fails, so tox's output is actually readable (otherwise I had to scroll through to find if the tests themselves had failed, or merely coveralls).

hobarrera:
I also added python3.5, which seems to be missing, to the dependency matrix.

bittner:
Thank you. Please, also update the [Travis configuration](https://github.com/jcassee/django-analytical/blob/master/.travis.yml) accordingly. See the comment at the beginning of the block.

The Python/Django combinations must match the officially supported versions (e.g. as [listed on PyPI](https://pypi.python.org/pypi/Django/1.9rc1)).

bittner:
Django 1.9 has been [released on Dec 1](https://www.djangoproject.com/weblog/2015/dec/01/django-19-released/).

@hobarrera Do you want to finish your pull request so we can merge it?

hobarrera:
Updated. I'm wanted to see if it was possible to do this programatically, rather than copy-pasting, but I honestly don't care enough about travis to investigate.

tox.ini
```diff
@@ -24,3 +18,4 @@ deps =
     django17: Django>=1.7,<1.8
     django18: Django>=1.8,<1.9
```
bittner: I'm wondering why Django 1.9 is installed though it's not defined under the dependencies. Probably the `Django` package is pulled without a version restriction.

Can you add a `django19` line, please? Thank you.

bittner:
Some builds fail. One reason is the missing Python 3.5 interpreter. This can be fixed as described in [Travis-CI issue 4794](https://github.com/travis-ci/travis-ci/issues/4794#issuecomment-143758799). I'd suggest to add the following line in `.travis.yml`, line 2:
```
python: 3.5
```
(this is, right after the `language: python` line)

tox.ini
hobarrera: Good catch, it seemed to be "working by coincidence" :)

bittner:
As for the `AppRegistryNotReady`, I've found this in the Django 1.7 release notes: "[in a plain Python script ... you must now explicitly initialize Django](https://docs.djangoproject.com/en/1.9/releases/1.7/#app-loading-changes)". `setup.py test` seems to fall under the "standalone script" category. But why didn't we hit this problem earlier? We have 1.7 in the test matrix for a while.

@jcassee Joost, do you see what's happening here?

hobarrera:
> As for the AppRegistryNotReady, I've found this in the Django 1.7 release notes: "in a plain Python script ... you must now explicitly initialize Django". setup.py test seems to fall under the "standalone script" category. But why didn't we hit this problem earlier? We have 1.7 in the test matrix for a while.

Regarding this, I'm not sure why use setup.py test, instead of django-admin test.

hobarrera:
... Actually, I've found that some tests are failing (not sure why they worked before; confusing).

I don't quite follow how the tests works; I've always used django's test framework for testing django app, and I'm not sure how `setup.py` interacts with it. I'll get back to you if I can't figure it out.

hobarrera:
> ... Actually, I've found that some tests are failing (not sure why they worked before; confusing).

Figured out why; not specifying the django version for the tox env `django19`, resulted in the latest stable, which was 1.8 at the time. So there's some stuff that needs fixing.

hobarrera:
Yeah, I've a single test failing on django1.9, but several now failing on older versions of django.
I've used django<1.7 almost nothing, so I'm really not familiar with lot of the internals.

I've almost used python3 mostly during the last 5 years, so I don't quite clearly know the differences to why some tests aren't getting picked up.

I pushed the latest of what I have, but I don't quite know how to:
(a) fix the test without breaking <1.6
(b) unbreak tests for older version.

IMHO, using django's test framework would remove a lot of pain, but I'm hesitant to try to write one that'll work okay on 1.4.
Unless there's some hurry for this, I'm willing to pick that up though.

jcassee:
I'm not sure what is going on here, but since this month Django 1.7 reached end of life. How about we trim the number of supported versions? 1.8 is another LTS release, so I would not mind using that as the lowest officially supported version.

hobarrera:
Well, if you're okay dropping anything below 1.7, I can re-write the unit tests to use django's framework, which should also give us less pain in future versions.

It'll be a bit of a big change, but should clean up quite a bit. :)

jcassee:
@bittner Your opinion?

bittner:
I'm unbiased with regard to this matter. I've always seen it as a specific quality of this app to support a large number of older Django releases. But of course, anyone with an old Django project can still use version 1.0 of django-analytical.

analytical/tests/test_tag_chartbeat.py
```diff
@@ -4,6 +4,7 @@
 
 import re
 
+from django import version
```
bittner: This line causes the issues. If you use try-except below a couple of failing builds will vanish.

hobarrera:
I've updated the PR. django1.9 now works fine, dropping unsupported django version.

hobarrera:
There's still an issue not caught by the test units. :(

Looks like `analytical.utils` imports `django.contrib.sites.models.Site`, which makes any apps that don't use `contrib.sites` fail. This emitted warning due to a deprecation since django 1.7, and has worked merely by coincidence so far.

I'll get back to you guys after I have a fix for that.

setup.py
```diff
@@ -47,6 +47,23 @@ def read(fname):
...
+    django.setup()
+except ImportError:
+    print(
+        "Could not import django."
+        "This is fine, unless you intend to run unit tests"
+    )
+except AttributeError:
+    # django.setup() does not exist on older djangos, so don't worry about
+    # that.
+    pass
```
bittner: Good. Though, actually, I think it should better fail (e.g. explicitly via `UnsupportedVersionError('Django < 1.7 not supported')`) instead of silencing it alltogether. As a goody the comment would then be unneeded.

A comment in Python usually reveals that there's something not as explicit or obvious as it should be.

tox.ini
```diff
@@ -1,26 +1,16 @@
...
-    py33: python3.3
-    py34: python3.4
+    sh -c 'coveralls | true'
 deps =
     coverage==3.7.1
     coveralls
-    django14: Django>=1.4,<1.5
-    django15: Django>=1.5,<1.6
-    django16: Django>=1.6,<1.7
-    django17: Django>=1.7,<1.8
```
bittner: I'd prefer to leave 1.7 in the supported versions, if it doesn't cause any additional trouble.

.travis.yml
```diff
@@ -10,8 +10,6 @@ env:  # generate list with: $ tox -l | xargs -I ITEM echo "  - TOXENV="ITEM
   - TOXENV=py27-django14
```
bittner: Please use the command mentioned in the comment above to generate an env value list matching `tox.ini`:
```
$ tox -l | xargs -I ITEM echo "  - TOXENV="ITEM
```

setup.py
hobarrera: You're right. This change was before dropping django < 1.7.

There is no `UnsupportedVersionError` in python, and I'd rather just get rid of the exception and raise that - in any case, supported versions should be well documented.

tox.ini
hobarrera: App initialization works differently, so the updated tests fail with django1.7, regrettably already. Since dj17 is unsupported already, and dj18 is an LTS, I don't think it'll hurt in the long run.

.travis.yml
hobarrera: Right, I keep forgetting about that. I'd leave an issue open to fix that permanently, since it's a bit of a hassle to have to keep that up-to-date (all tests pass locally, BTW).

The configuration for this repo might serve as a reference:
 * [travis config](https://github.com/geier/khal/blob/master/.travis.yml)
 * [travis script](https://github.com/geier/khal/blob/master/.travis.run.sh)

hobarrera:
Bump? :D

jcassee:
I don't mind dropping support for 1.7 if supporting it would be awkward. @bittner: do you see a way of maintaining compatibility?

bittner:
I don't understand "App initialization works differently". @hobarrera Can you point to a line of code that's causing incompatibility?

hobarrera:
> I don't understand "App initialization works differently". @hobarrera Can you point to a line of code that's causing incompatibility?

Trying to pinpoint *exactly* what line caused the issue and what it was, I managed not only find it, but also fix it. Django1.7 now works.

All tests now pass with django>=1.7.

hobarrera:
ping?

jcassee:
Strange, the Python 3.5 interpreter cannot be found, but Travis-CI [says they do support it](https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against)? I will ask them about it.

bittner:
@jcassee Joost, look at the documentation. I think I provided links above.
We have to specify the python version as 3.5 in the Travis file. Otherwise
python3.5 is not yet available in the VM image.

Peter

Am Mittwoch, 16. Dezember 2015 schrieb Joost Cassee :
> Strange, the Python 3.5 interpreter cannot be found, but Travis-CI says
they do support it? I will ask them about it.

jcassee:
Good point, let's add it to `.travis.yml` then and try again. :smile:

jcassee:
@bittner Ehm, if I add versions to the `python` list, won't that run the tests multiple times? We already have that covered by `env`.

The docs say "For a limited number of Python development releases, on-demand installation is available. Currently, these are: 3.5-dev (built nightly), 3.5/3.5.0.". No idea what that means, though.

hobarrera:
Sorry, does this need some intervention on my behalf, or is that just a travis configuration that needs to be done by the account owner?

Tests are all passing right now (with all interpreters) when running locally.

jcassee:
@hobarrera If you change `.travis.yml` in the pull request, Travis will run the tests with the new configuration. So if you have the time, please contribute! (I was waiting for a reply to my request mail to Travis support.)

bittner:
@jcassee No, Joost, we just need to have `python: 3.5` in the Travis file (= no multiple times). See my [comment above](https://github.com/jcassee/django-analytical/pull/73#issuecomment-162206854), the link goes to an example.

jcassee:
Ah, okay. I don't think I can push to this branch, so I will let @hobarrera add the change.

hobarrera:
Travis is now passing all tests as well. 😃

bittner:
Yeah, let's go party!   :tada:   Good job!

.travis.yml
```diff
@@ -1,24 +1,19 @@
...
-  - TOXENV=py32-django15
-  - TOXENV=py32-django16
   - TOXENV=py32-django17
-  - TOXENV=py32-django18
-  - TOXENV=py33-django15
-  - TOXENV=py33-django16
   - TOXENV=py33-django17
-  - TOXENV=py33-django18
   - TOXENV=py34-django17
+  - TOXENV=py33-django18
```
bittner: The order is a bit mixed up here. Can you sort the list by Python version first, Django version next?

bittner:
A last thing:

The README and/or the documentation should better mention that we're dropping support for Django<1.7, and why we do this. Can you add this in an appropriate place, @hobarrera?

Cheers! Awesome job so far!

.travis.yml
hobarrera: This is actually auto-generated code, so I don't think that really matters. I'd love to see this disappear in favour of some programatical approach that does not require manually updating this section.

khal (totally unrelated) deals nicely with this:

https://github.com/geier/khal/blob/master/.travis.yml
https://github.com/geier/khal/blob/master/.travis.run.sh

hobarrera:
Added it to the README (and the docs/index).

jcassee:
Alright, time to merge, right @bittner?

setup.py
```diff
@@ -47,6 +47,19 @@ def read(fname):
...
+    import django
+    os.environ.setdefault(
+        "DJANGO_SETTINGS_MODULE",
+        "analytical.tests.settings"
+    )
+    django.setup()
+except ImportError:
+    print(
+        "Could not import django."
+        "This is fine, unless you intend to run unit tests"
```
bittner: There's at least a blank missing between the two sentences. They will stick together otherwise. Maybe also a final period could be added.

.travis.yml
bittner: @hobarrera Are you sure this is autogenerated code? According to the modified `tox.ini` there would be `py27-*` entries following after the last `py35-*`.

I don't want to be picky, but I feel we should care about consistency. Either we make this code here match the output of `tox -l` exactly (that's why there is the comment at the beginning of the `env` block after all), or at least we make the list beautiful in an obvious way.

Suggestion: Please change the comment on top of the `env` block to:
````bash
$ tox -l | sort | xargs -I ITEM echo "  - TOXENV="ITEM
```
(Note the additional, piped `sort` call.) Then generate the list again with that command, and paste it in here.

P.S.: I don't think it's easy to "automate" this in a sensible way, because we have 2 dimensions (python, django). If it is only one, it's fairly straightforward with Travis-CI. See [setup.cfg](https://github.com/Organice/django-organice/blob/master/setup.cfg#L17) and [.travis.yml](https://github.com/Organice/django-organice/blob/master/.travis.yml#L12) in django-organice, for example. -- If you can, though, do it. That's appreciated. I don't like crypographic shell scripting though. As we say in Python, *"There must be a better way!"*

README.rst
```diff
@@ -21,6 +21,9 @@ set-up very simple, while allowing advanced users to customize tracking.
 Each service is set up as recommended by the services themselves, using
 an asynchronous version of the Javascript code if possible.
 
+Version 1.0.0 is the last to support Django < 1.7. Users of older django
+versions should stick to 1.0.0, and are encouraged to upgrade their setups.
+
```
bittner: We also need to update the Changelog in the documentation. @jcassee Will you do this?

hobarrera:
bump?

bittner:
@hobarrera There are the two comments ([1](https://github.com/jcassee/django-analytical/pull/73#discussion_r48021652), [2](https://github.com/jcassee/django-analytical/pull/73#discussion_r48020725)) from 3 days ago that I think should still be addressed. That's what I saw. The rest is up to Jost @jcassee.

.travis.yml
hobarrera: I've updated `.travis.yml` with both the sorting, and the generation code.

> P.S.: I don't think it's easy to "automate" this in a sensible way

I've posted a link to a scenario where this has been achieved already, with just a couple of lines.

> I don't like crypographic shell scripting though.

I linked to a very clear shell script that deals with this. A simple script is infinitely superior to having to remind devs to manually update files after committing certain files.

We can, of course, rewrite it into python, if you like, though it seems a bit of an overkill for just a one-line `sed`.

setup.py
hobarrera: Well spotted. Fixed.

hobarrera:
@bittner Oops, I kind of thought I had addressed those. Updated PR now. :)

bittner:
Happy New Year! -- Ready to roll, @jcassee?

jcassee:
Yes! I just merged another PR which causes a conflict in this one. I will resolve it and merge.

hobarrera:
> Yes! I just merged another PR which causes a conflict in this one. I will resolve it and merge.

I guess you did it yourself, since I'm seeing this merge into master, right? :)

Thanks.

PR #74, opened 2015-12-17
override internal ips with empty list

brad:
I wanted to be able to set `ANALYTICAL_INTERNAL_IPS=[]` on my development machine, however when the list is empty it doesn't properly override what is in `INTERNAL_IPS`. This PR aims to fix that.


jcassee:
Thanks, Brad!

PR #79, opened 2016-01-17
Move comment on Django support from README to history

bittner:
I believe the paragraph on Django support is better placed in the History chapter of the documentation. Makes the README unnecessarily verbose.

BTW, I adore your clean structure (allowing both self-explaining files in the project root and inclusion in the docs) with AUTHORS, CHANGELOG, LICENSE.


jcassee:
Agreed!

Issue #83, opened 2016-04-03
Support Yandex.Metrica


sikmir:
[Yandex.Metrica](https://metrica.yandex.com) is an analytics tool like as google analytics.

bittner:
Fixed via #84.

PR #84, opened 2016-04-03
Support Yandex.Metrica

sikmir:
Issue #83

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,7 @@ Currently Supported Services
 * `Spring Metrics`_ conversion tracking
 * `UserVoice`_ user feedback and helpdesk
 * `Woopra`_ web analytics
+* `Yandex.Metrica`_ web analytics
 
 .. _`Chartbeat`: http://www.chartbeat.com/
 .. _`Clickmap`: http://getclickmap.com/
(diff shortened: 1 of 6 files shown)


bittner:
@jcassee Looks like there is an issue with `pip` version 8.1.1 for [builds running Python 3.2](https://travis-ci.org/jcassee/django-analytical/jobs/120400903#L212), most probably unrelated to the code of this PR.

[Line 204 of the log](https://travis-ci.org/jcassee/django-analytical/jobs/120400903#L204) says: "UserWarning: Support for Python 3.0-3.2 has been dropped. Future versions will fail here." -- We have to pin / install an earlier version of `pip`, or drop Python 3.2 support altogether.

jcassee:
I would not mind dropping support for Python 3.0-3.2. All distributions that come with Python 3 are on >= 3.3, right?

jcassee:
:+1: Changes look good.

bittner:
Very bad: Django 1.8 still runs on Python 3.2. But I guess we're not the only ones being affected. I would first try to pin the version of `pip<8.1.1`, though.

jcassee:
Ah, we need to support Python 3.2 then. Let's try pinning pip then.

jcassee:
@sikmir Could you rebase against the new master branch? This should trigger a rebuild that should fix the failing checks.

bittner:
Same here as in [PR #88](https://github.com/jcassee/django-analytical/pull/88#issuecomment-220002305). Also the coverage decreased and needs action.

@jcassee What's the best place to add a test for triggering the [TemplateSyntaxError](https://coveralls.io/jobs/14567773/source_files/496089754#L53)?

jcassee:
@bittner Probably `test_tag_analytical.py`?

bittner:
I think we can merge this PR. What do you think, @jcassee?

PR #86, opened 2016-04-17
Add "Python versions" badge (README)

bittner:
Discovered this today. Very sweet!

PR #89, opened 2016-05-17
Google Analytics _set*SampleRate and _set*CookieTimeout

saschwarz:
I added support for GA's `_setSampleRate` and `_setSiteSpeedSampleRate` settings for my needs and then saw `_setSessionCookieTimeout` and `_setVisitorCookieTimeout` would be easy to add too.

I've updated the tests and the docs and thought it might be of interest to other developers.
Thanks for this project!
Steve

### DIFF
--- a/analytical/templatetags/google_analytics.py
+++ b/analytical/templatetags/google_analytics.py
@@ -4,6 +4,7 @@ Google Analytics template tags and filters.
 
 from __future__ import absolute_import
 
+import decimal
 import re
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


bittner:
Hi @saschwarz, thanks for your contribution! Just wanted to say, don't worry about the failing build: That's just for Python 3.2, and will go away once #86 is merged.

analytical/templatetags/google_analytics.py
```diff
@@ -131,8 +139,37 @@ def _get_other_commands(self, context):
...
         if getattr(settings, 'GOOGLE_ANALYTICS_SITE_SPEED', False):
             commands.append(SITE_SPEED_CODE)
+
         if getattr(settings, 'GOOGLE_ANALYTICS_ANONYMIZE_IP', False):
             commands.append(ANONYMIZE_IP_CODE)
+
+        sampleRate = getattr(settings, 'GOOGLE_ANALYTICS_SET_SAMPLE_RATE', False)
+        if sampleRate is not False:
+            value = decimal.Decimal(sampleRate)
+            if value < 0 or value > 100:
```
bittner: Could be written as:
``` python
            if not 0 <= value <= 100:
```

analytical/templatetags/google_analytics.py
```diff
@@ -131,8 +139,37 @@ def _get_other_commands(self, context):
...
+                raise AnalyticalException("'GOOGLE_ANALYTICS_SET_SAMPLE_RATE' must be >= 0 and <= 100")
+            commands.append(SET_SAMPLE_RATE_CODE % value.quantize(TWOPLACES))
+
+        siteSpeedSampleRate = getattr(settings, 'GOOGLE_ANALYTICS_SET_SITE_SPEED_SAMPLE_RATE', False)
+        if siteSpeedSampleRate is not False:
+            value = decimal.Decimal(siteSpeedSampleRate)
+            if value < 0 or value > 100:
```
bittner: Could be written as:
``` python
            if not 0 <= value <= 100:
```

jcassee:
I have a question: why do the settings have `_SET_` in them? I would have expected names like `GOOGLE_ANALYTICS_SAMPLE_RATE`. Do you feel it adds something?

saschwarz:
@jcassee I was just "underscoring" the names GA uses in their variable names: '_setSampleRate' -> 'SET_SAMPLE_RATE'. But you've made me realize I didn't follow the precedent set in the existing tags. I can update the PR if you like.

bittner:
@saschwarz If you merge `master` into your feature branch (or rebase to HEAD) the build will fully pass.

jcassee:
If possible I would prefer rebasing.

saschwarz:
@jcassee would you like the settings renamed? I can do that tonight and then squash/rebase.

jcassee:
Yeah, unless you have an argument against. It's four characters less to type. 😄

saschwarz:
The majority begin with `_set` . Notable exceptions appear to be for ecommerce/event tracking but those are more likely to be invoked dynamically by JS on the client.

bittner:
I also vote for removing `SET_` from the constants. Just because Google engineers write questionable code this doesn't mean we should forget [PEP 20](http://legacy.python.org/dev/peps/pep-0020/), the Zen of Python. Let's put readable code over compliance with Google. _"Beautiful is better than ugly"_

PR #95, opened 2016-05-27
GA flags with incorrect tags

saschwarz:
The *SAMPLE_RATE* and *COOKIE_TIMEOUT* keys were incorrectly pushing the '_gat.' flag into the GA configuration array.


jcassee:
Released as 2.2.1

PR #97, opened 2016-09-06
Update intercom id format

silsha:
The new ids are much shorter (8 characters) can include all lowercase letters and numbers.

### DIFF
--- a/analytical/templatetags/intercom.py
+++ b/analytical/templatetags/intercom.py
@@ -12,7 +12,7 @@ from django.template import Library, Node, TemplateSyntaxError
 from analytical.utils import disable_html, get_required_setting, \
         is_internal_ip, get_user_from_context, get_identity
 
-APP_ID_RE = re.compile(r'[\da-f]+$')
+APP_ID_RE = re.compile(r'[\da-z]+$')
 TRACKING_CODE = """
 <script id="IntercomSettingsScriptTag">
   window.intercomSettings = %(settings_json)s;


bittner:
Hi @silsha, thank you for placing a PR for django-analytical!

- Do you have a reference for the change? I can't find any in the [Intercom docs](https://docs.intercom.io/install-on-your-product-or-site/quick-install/install-intercom-on-your-web-app).
- Would it make sense to adapt [our tests](https://github.com/jcassee/django-analytical/blob/master/analytical/tests/test_tag_intercom.py#L16) for the Intercom integration?

analytical/templatetags/intercom.py
```diff
@@ -12,7 +12,7 @@
 from analytical.utils import disable_html, get_required_setting, \
         is_internal_ip, get_user_from_context, get_identity
 
-APP_ID_RE = re.compile(r'[\da-f]+$')
+APP_ID_RE = re.compile(r'[\da-z]+$')
```
bittner: Please update the test regarding this change.

bittner:
Interesting that [no existing test](https://github.com/jcassee/django-analytical/blob/master/analytical/tests/test_tag_intercom.py#L44-L50) detects the changes.

@silsha Do you care to add a test that ensures that the regex is safeguarding the integration? We could then merge your changes.

bittner:
@silsha Would you be interested in taking this PR forward?

Vitiell0:
Please publish this change. I just spent a hour figuring this out myself as I was trying to get intercom working with analytical.

bittner:
@Vitiell0 If you would dare to update the tests for this change we could publish it altogether.

It probably makes sense to make a new PR for this. I fear @silsha has abandoned this PR. :disappointed:

Vitiell0:
Ok I'll give it a shot and create a new PR

bittner:
Awesome! - Try to make the tests fail your previous, broken setup. Then fix your setup, and the tests should pass. That would be really cool.

jcassee:
Superseded by #105.

PR #98, opened 2016-09-30
Added optional port to Piwik domain path

ianalexr:
Small addition allowing a domain port with the Piwik module's domain regex. Ran into this when setting up django-analytical and Piwik for a project.


jcassee:
Thanks Alex! I will release a new version with your fix shortly.

PR #101, opened 2017-01-12
Use protocol-relative URL for Piwik fallback img

### LINKED ISSUE #96: Piwik fallback img tracking is only served as plain HTTP
In the `noscript` case, the tracking img is served as HTTP only:
https://github.com/jcassee/django-analytical/blob/master/analytical/templatetags/piwik.py#L37

However, if the site is served through HTTPS, this might trigger security warnings.

As a side note, since this is a `templatetag`, this could use request context to detect scheme and not rely on javascript to switch between HTTP and HTTPS.

If this makes sense to you, I'll provide a patch.

garrettr:
Avoids mixed content issues when using django-analytical on sites served
over HTTPS. Using a protocol-relative URL is also recommended in the
Piwik documentation: https://issues.piwik.org/344.

Fixes #96.

### DIFF
--- a/analytical/templatetags/piwik.py
+++ b/analytical/templatetags/piwik.py
@@ -21,7 +21,6 @@ DOMAINPATH_RE = re.compile(r'^(([^./?#@:]+\.)*[^./?#@:]+)+(:[0-9]+)?(/[^/?#@:]+)
 SITEID_RE = re.compile(r'^\d+$')
 
 TRACKING_CODE = """
-<!-- Piwik -->
 <script type="text/javascript">
   var _paq = _paq || [];
   %(variables)s
@@ -35,7 +34,6 @@ TRACKING_CODE = """
     g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
   })();
 </script>
-<!-- End Piwik Code -->
 <noscript><p><img src="//%(url)s/piwik.php?idsite=%(siteid)s" style="border:0;" alt="" /></p></noscript>
 """  # noqa
 
--- a/analytical/templatetags/piwik.py
+++ b/analytical/templatetags/piwik.py
@@ -21,19 +21,21 @@ DOMAINPATH_RE = re.compile(r'^(([^./?#@:]+\.)*[^./?#@:]+)+(:[0-9]+)?(/[^/?#@:]+)
 SITEID_RE = re.compile(r'^\d+$')
 
 TRACKING_CODE = """
+<!-- Piwik -->
 <script type="text/javascript">
   var _paq = _paq || [];
   %(variables)s
   _paq.push(['trackPageView']);
   _paq.push(['enableLinkTracking']);
   (function() {
-    var u=(("https:" == document.location.protocol) ? "https" : "http") + "://%(url)s/";
+    var u="//%(url)s/";
     _paq.push(['setTrackerUrl', u+'piwik.php']);
     _paq.push(['setSiteId', %(siteid)s]);
-    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0]; g.type='text/javascript';
-    g.defer=true; g.async=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
   })();
 </script>
+<!-- End Piwik Code -->
 <noscript><p><img src="//%(url)s/piwik.php?idsite=%(siteid)s" style="border:0;" alt="" /></p></noscript>
 """  # noqa
 
--- a/analytical/templatetags/piwik.py
+++ b/analytical/templatetags/piwik.py
@@ -34,7 +34,7 @@ TRACKING_CODE = """
     g.defer=true; g.async=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
   })();
 </script>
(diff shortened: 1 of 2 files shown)


garrettr:
In making this PR, I noticed that the [JS tracking code](https://github.com/jcassee/django-analytical/blob/d9b482b8c70a93ac377e2906af071be1e05bf893/analytical/templatetags/piwik.py#L23-L38) embedded for Piwik by django-analytical does not match the tracking code in the [latest Piwik docs](https://developer.piwik.org/guides/tracking-javascript-guide). The Piwik docs also say:

> If your Piwik tracking code doesn't look like this one, you may be using the deprecated version. Older versions still work as expected and will track your visitors, but we highly recommend that you update your pages to use the most recent tracking code.

Would it be helpful for me to file an additional PR that updates django-analytical to use the most recent tracking code for Piwik?

bittner:
I don't mind if you make the changes to the Piwik JavaScript snippet also in this PR.

Also, if you would address [my recommendation](https://github.com/jcassee/django-analytical/issues/96#issuecomment-266358140) re `PIWIK_URL` vs. `PIWIK_DOMAIN_PATH` that would be very nice.

garrettr:
> Also, if you would address my recommendation re PIWIK_URL vs. PIWIK_DOMAIN_PATH that would be very nice.

I decided that I didn't need to do that because I did not implement my earlier [proposal](https://github.com/jcassee/django-analytical/issues/96#issuecomment-266149194) to allow PIWIK_DOMAIN_PATH to include the scheme. If you want me to implement PIWIK_URL, I could, but after further reflection I simply don't think it's necessary. Simply using protocol-relative URLs everywhere avoids mixed content issues in every context, which is the most important thing to fix.

analytical/templatetags/piwik.py
```diff
@@ -21,20 +21,22 @@
 SITEID_RE = re.compile(r'^\d+$')
 
 TRACKING_CODE = """
+<!-- Piwik -->
```
bittner: I see this is in the [official Piwik sample](https://developer.piwik.org/guides/tracking-javascript-guide). For me, comments in HTML are a bad habit (from a time when PHP was the awesome child). If we need comments at all we can use Python comments or a comment tag in the template.

Let's not include HTML comments in our code. (Unless we really need help to analyze our HTML output.)

analytical/templatetags/piwik.py
```diff
@@ -21,20 +21,22 @@
...
     _paq.push(['setTrackerUrl', u+'piwik.php']);
     _paq.push(['setSiteId', %(siteid)s]);
-    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0]; g.type='text/javascript';
-    g.defer=true; g.async=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
   })();
 </script>
-<noscript><p><img src="http://%(url)s/piwik.php?idsite=%(siteid)s" style="border:0;" alt="" /></p></noscript>
+<!-- End Piwik Code -->
```
bittner: Same here: Let's not include the (optional) HTML comments from the Piwik snippet.

analytical/tests/test_tag_piwik.py
```diff
@@ -20,37 +20,37 @@ class PiwikTagTestCase(TagTestCase):
...
+        self.assertTrue('img src="//example.com/piwik.php?idsite=345"'
                         in r, r)
 
     @override_settings(PIWIK_DOMAIN_PATH='example.com/piwik',
                        PIWIK_SITE_ID='345')
     def test_domain_path_valid(self):
         r = self.render_tag('piwik', 'piwik')
-        self.assertTrue(' ? "https" : "http") + "://example.com/piwik/";' in r,
+        self.assertTrue('"//example.com/piwik/"' in r,
                         r)
```
bittner: Please join the `r)` with the line above. Now it fits the 80 characters limit.

analytical/tests/test_tag_piwik.py
```diff
@@ -20,37 +20,37 @@ class PiwikTagTestCase(TagTestCase):
...
 
     @override_settings(PIWIK_DOMAIN_PATH='example.com:1234',
                        PIWIK_SITE_ID='345')
     def test_domain_port_valid(self):
         r = self.render_tag('piwik', 'piwik')
-        self.assertTrue(' ? "https" : "http") + "://example.com:1234/";' in r,
+        self.assertTrue('"//example.com:1234/";' in r,
                         r)
```
bittner: Same here: Please join the `r)` with the line above. Now it fits the 80 characters limit.

analytical/tests/test_tag_piwik.py
```diff
@@ -20,37 +20,37 @@ class PiwikTagTestCase(TagTestCase):
...
 
     @override_settings(PIWIK_DOMAIN_PATH='example.com:1234/piwik',
                        PIWIK_SITE_ID='345')
     def test_domain_port_path_valid(self):
         r = self.render_tag('piwik', 'piwik')
-        self.assertTrue(' ? "https" : "http") + "://example.com:1234/piwik/";' in r,
+        self.assertTrue('"//example.com:1234/piwik/"' in r,
                         r)
```
bittner: And again: Please join the `r)` with the line above. Now it fits the 80 characters limit.

garrettr:
@bittner What's the status on this? Are there any further changes you'd like me to make before this is ready to merge?

bittner:
@jcassee Can you take a look at the changes, too, please?

garrettr:
@jcassee Any thoughts on if this might make it into an official release sometime in the near future?

jcassee:
Sure, I'll release a new version shortly.

PR #105, opened 2017-02-15
Updated regex for Intercom APP_ID and tests for Intercom

Vitiell0:
Intercom has updated their ID to include characters a-z.


bittner:
Looks good to me.

@jcassee Can you take a look at the tests? Do they still make sense? - I'm especially worried about why [none of the existing tests detected the code changes](https://github.com/jcassee/django-analytical/pull/97#issuecomment-274189645), as commented earlier this year.

jcassee:
Change looks good, thanks @Vitiell0!

@bittner Well, I think the test for a correct ID still works because `[a-z]` includes `[a-f]`, and the test for an incorrect ID still works because it tested for a different "incorrectness": a capital letter. It is difficult to define a boundary in this case.

Issue #104, opened 2017-01-29
Extend the tests (tox, Travis CI) to Django 1.10


bittner:
Extending the tests to Django 1.10 is more than simply updating [tox.ini](https://github.com/jcassee/django-analytical/blob/master/tox.ini#L5) and [.travis.yml](https://github.com/jcassee/django-analytical/blob/master/.travis.yml#L23): A DjangoTemplates backend needs to be configured. A tox run yields errors for Django 1.10:

```
django.core.exceptions.ImproperlyConfigured: No DjangoTemplates backend is configured.
```

Related Django docs
---------------

- [Topics: Templates](https://docs.djangoproject.com/en/1.10/topics/templates/#configuration)
- [Reference: Templates](https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/#the-templates-settings)  (upgrading)

pjdelport:
#114 should fix this.

jcassee:
Fixed by #144

PR #114, opened 2017-10-17
Update tests for Django 1.10 & 1.11

pjdelport:
This updates the test settings and Tox / Travis configs to support Django 1.10 and 1.11 (as well as Python 3.6, where relevant).


pjdelport:
This should resolve #104.

tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
 envlist =
+    # Django 1.7 does not support Python 3.5+
     py{27,33,34}-django17
-    py{27,33,34,35}-django18
-    py{27,34,35}-django19
-#    py{27,34,35}-django110
-#    py{27,34,35,36}-django111
+    # Django 1.9+ does not support Python 3.3
```
bittner: You don't need to add those comments. We would also have to mention Python 3.2, which [Django 1.7](https://pypi.python.org/pypi/Django/1.7.11) would support but is excluded from the build matrix from sane reasons.

Please keep a continuous list of Djangos, from `django17` down to `django111` with the Python versions specified with expansion syntax. Thanks!

.travis.yml
```diff
@@ -4,19 +4,27 @@ python:
...
+  - DJANGO=1.10
+  - DJANGO=1.11
 matrix:
   exclude:
+    # Django 1.7 does not support Python 3.5+
     - python: 3.5
       env: DJANGO=1.7
+    - python: 3.6
+      env: DJANGO=1.7
+    # Django 1.9+ does not support Python 3.3
```
bittner: You don't need to add those comments. The data should be self-explanatory (`exclude` section. Python + Django versions specified). -- Comments are usually an indicator that the code or configuration is not self-explantory.

.travis.yml
```diff
@@ -4,19 +4,27 @@ python:
...
     - python: 3.3
       env: DJANGO=1.9
+    - python: 3.3
+      env: DJANGO=1.10
+    - python: 3.3
+      env: DJANGO=1.11
   include:
```
bittner: Please, also add the exceptions for Python 3.6. I think the whole list should look like this:
```yml
matrix:
  exclude:
    - python: 3.5
      env: DJANGO=1.7
    - python: 3.6
      env: DJANGO=1.7
    - python: 3.6
      env: DJANGO=1.8
    - python: 3.3
      env: DJANGO=1.9
    - python: 3.6
      env: DJANGO=1.9
    - python: 3.3
      env: DJANGO=1.10
    - python: 3.6
      env: DJANGO=1.10
    - python: 3.3
      env: DJANGO=1.11
```

.travis.yml
pjdelport: Hmm, it's not self-explanatory to me: I had to go look up on the Django docs which combinations are actually supported (which is why I added the comments).

.travis.yml
pjdelport: I'm not sure I understand why Python 3.6 should be excluded: it works in production, and all the tests pass. Surely it's useful to test on the latest Pythons for the users that use it (myself included)?

.travis.yml
bittner: Well, that's normal. Actually, the docs could err. I prefer to look at the PyPI package meta data.

I would prefer to keep the Travis configuration clean. As I said, you would otherwise have to explain why Python 3.2 is missing.

bittner: I don't want to be the dog in the manger:

If you really care that much you may add a general comment above the `exclude:` section, kind of "Exclude unsupported Python/Django combinations", or so. If it's really necessary, in the end.

.travis.yml
bittner: > I'm not sure I understand why Python 3.6 should be excluded

It's not our fault. Django versions below 1.11 don't officially support 3.6, that's it.

You can [browse throught the Django releases](https://pypi.python.org/pypi/Django) on PyPI and look at the meta data (trove classifiers). Python 3.6 is supported since [Django 1.11](https://pypi.python.org/pypi/Django/1.11).

.travis.yml
jcassee: The sections specify the versions, so I think one comment is exactly the right number; no comment and I begin to wonder why we exclude these. This does not block the build for me, though.

.travis.yml
pjdelport: Right, I know they're not officially supported; but neither are Python 3.3 or Django 1.7 & 1.9, and the tests still cover those, because it's still useful for the people who use them.

Python is generally expected to be backward compatible, so it should be very reasonable to always test on all forward versions of Python that are known to work, and only exclude the ones that are known to be broken, and won't be fixed (like Django 1.7 and Python 3.5+).

.travis.yml
pjdelport: @jcassee: The reason I made both comments is to explicitly document the different version ranges that are known to be broken (only Django 1.7 with any Python 3.5+, versus only Python 3.3 with any Django 1.9+).

Having only a list of existing version numbers doesn't make it obvious what ranges they are intended to represent, and doesn't give any guidance on how exactly how new Django and Python releases should be added to the list, or not. The comments hopefully make this more obvious. (They're certainly useful to me, for future maintenance.)

.travis.yml
```diff
@@ -4,22 +4,24 @@ python:
...
+    - { python: 3.5, env: DJANGO=1.7 }
+    - { python: 3.6, env: DJANGO=1.7 }
+    # Django 1.9+ does not support Python 3.3
+    - { python: 3.3, env: DJANGO=1.9 }
+    - { python: 3.3, env: DJANGO=1.10 }
+    - { python: 3.3, env: DJANGO=1.11 }
   include:
-    - python: 2.7
-      env: TOXENV=flake8
+    - { python: 2.7, env: TOXENV=flake8 }
```
pjdelport: I compacted these so that they're one per line, and hopefully more readable. Does that look good?

jcassee:
LGTM, @bittner?

.travis.yml
bittner: Not bad! Though my eyes feel irritated with curly braces in a YAML file!  :nerd_face: 

I'm still shocked that you care so much about those comments. Here comes my okay: :+1: :confounded:

bittner: Maybe it's not all that bad ... :smiley: ... in combination with the quirky curly braces!

Thanks for not giving up on us!  :1st_place_medal:

bittner:
Yes, LGTM! -- WETM! (whatever this means)

:+1:

bittner:
Ah! _"Looks good to me"_

Sorry, took me some time...

jcassee:
Released in version 2.3.0. Thanks, @pjdelport!

PR #115, opened 2017-10-17
Add Facebook Pixel support

pjdelport:
This adds support for Facebook Pixel, with tests and docs.

### DIFF
--- a/analytical/templatetags/facebook_pixel.py
+++ b/analytical/templatetags/facebook_pixel.py
@@ -1,0 +1,97 @@
+"""
+Facebook Pixel template tags and filters.
+"""
+from __future__ import absolute_import
+
+import re
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import get_required_setting, is_internal_ip, disable_html
+
+
+FACEBOOK_PIXEL_HEAD_CODE = """\
+<script>
+  !function(f,b,e,v,n,t,s)
+  {if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+  n.callMethod.apply(n,arguments):n.queue.push(arguments)};
+  if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+  n.queue=[];t=b.createElement(e);t.async=!0;
+  t.src=v;s=b.getElementsByTagName(e)[0];
+  s.parentNode.insertBefore(t,s)}(window, document,'script',
+  'https://connect.facebook.net/en_US/fbevents.js');
+  fbq('init', '%(FACEBOOK_PIXEL_ID)s');
+  fbq('track', 'PageView');
+</script>
+"""
+
+FACEBOOK_PIXEL_BODY_CODE = """\
+<noscript><img height="1" width="1" style="display:none"
+  src="https://www.facebook.com/tr?id=%(FACEBOOK_PIXEL_ID)s&ev=PageView&noscript=1"
+/></noscript>
+"""
+
+register = Library()
+
+
+def _validate_no_args(token):
+    bits = token.split_contents()
+    if len(bits) > 1:
+        raise TemplateSyntaxError("'%s' takes no arguments" % bits[0])
+
+
+@register.tag
+def facebook_pixel_head(parser, token):
+    """
+    Facebook Pixel head template tag.
+    """
+    _validate_no_args(token)
+    return FacebookPixelHeadNode()
+
+
+@register.tag
+def facebook_pixel_body(parser, token):
+    """
+    Facebook Pixel body template tag.
+    """
+    _validate_no_args(token)
+    return FacebookPixelBodyNode()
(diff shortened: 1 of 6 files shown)


pjdelport:
Is that failure just due to the average coverage percentage dropping?

analytical/tests/test_tag_facebook_pixel.py
```diff
@@ -0,0 +1,87 @@
...
+  t.src=v;s=b.getElementsByTagName(e)[0];
+  s.parentNode.insertBefore(t,s)}(window, document,'script',
+  'https://connect.facebook.net/en_US/fbevents.js');
+  fbq('init', '1234567890');
+  fbq('track', 'PageView');
+</script>
+"""
+
+
+expected_body_html = """\
```
bittner: Do we need the backslash?

analytical/tests/test_tag_facebook_pixel.py
```diff
@@ -0,0 +1,87 @@
...
+from django.http import HttpRequest
+from django.template import Context
+from django.test import override_settings
+
+from analytical.templatetags.facebook_pixel import FacebookPixelHeadNode, FacebookPixelBodyNode
+from analytical.tests.utils import TagTestCase
+from analytical.utils import AnalyticalException
+
+
+expected_head_html = """\
```
bittner: Is the backslash needed?

analytical/templatetags/facebook_pixel.py
```diff
@@ -0,0 +1,97 @@
...
+  n.queue=[];t=b.createElement(e);t.async=!0;
+  t.src=v;s=b.getElementsByTagName(e)[0];
+  s.parentNode.insertBefore(t,s)}(window, document,'script',
+  'https://connect.facebook.net/en_US/fbevents.js');
+  fbq('init', '%(FACEBOOK_PIXEL_ID)s');
+  fbq('track', 'PageView');
+</script>
+"""
+
+FACEBOOK_PIXEL_BODY_CODE = """\
```
bittner: Same here: Is the backslash needed? (`flake8` didn't complain, huh?!?)

analytical/templatetags/facebook_pixel.py
```diff
@@ -0,0 +1,97 @@
...
+from __future__ import absolute_import
+
+import re
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import get_required_setting, is_internal_ip, disable_html
+
+
+FACEBOOK_PIXEL_HEAD_CODE = """\
```
bittner: Backslash? (Do we have this in the other implementations, too?)

bittner:
> Is that failure just due to the average coverage percentage dropping?

Strange. You have added tests. @jcassee, can you explain that?

jcassee:
Yes, the failure is due to Coveralls. Ignore it, it's not accurate in this case.

analytical/tests/test_tag_facebook_pixel.py
pjdelport: This is the conventional way (as far as I know) to avoid inserting an unwanted extra empty line in the beginning of a block string literal. So for example:

```Python console
>>> """\
... One
... Two
... """.splitlines()
['One', 'Two']

>>> """
... One
... Two
... """.splitlines()
['', 'One', 'Two']
```

In this case, the purpose is just to have nice clean HTML output. :)

pjdelport:
The Coveralls report seems to say that coverage delta is due to the one extra line being added to `analytical/templatetags/analytical.py` (the only change in that file), even though it's part of a list literal, and not a line of code?

In other words, it's counting lines covered versus the raw line count of the file, rather than the number of coverage-eligible lines in the file? This seems to be a known issue: https://github.com/lemurheavy/coveralls-public/issues/636, https://github.com/lemurheavy/coveralls-public/issues/976

pjdelport:
I brought the `facebook_pixel` test coverage up to 100%, so the average coverage delta is now positive. \o/

analytical/tests/test_tag_facebook_pixel.py
jcassee: Indeed, this is convention, although I cannot find an authoritative link at the moment.

jcassee:
LGTM, @bittner?

analytical/tests/test_tag_facebook_pixel.py
pjdelport: [This StackOverflow question](https://stackoverflow.com/q/2504411/444705) has some examples of this pattern (also in conjunction with `dedent()` for indentation stripping).

bittner:
LGTM?? = Looks Good To Merge??  :smirk: :grey_question: 

Yes! :smiley:

jcassee:
Released in version 2.3.0. Thanks, @pjdelport!

pjdelport:
Thanks for releasing this so promptly! :D

PR #118, opened 2017-11-18
Check PyPI package description generation w/ tox

bittner:
Converting the README.rst to HTML sometimes fails on PyPI and causes unreadable package descriptions on PyPI. This PR adds a check to tox and Travis, which verifies that the package description on PyPI will be generated correctly.


jcassee:
Nice, @bittner, I've had that happen once or twice.

bittner:
Sure! One thing less to worry about.

I hope you're doing fine, Joost. Greetings from chilly lake Constance!

PR #119, opened 2017-11-23
Use more targeted approach for README conversion check

bittner:
With this change we use the somewhat dedicated approach for checking for a correct .rst -> .html conversion before packaging. Seen at https://github.com/jazzband/pip-tools/pull/608.


jcassee:
Was sick in bed the last week, so only looking at it now. Looks good! I'll just merge.

bittner:
Hope you're well again now. Have a great weekend!

Issue #120, opened 2017-12-02
Support Django 2.0


Anteru:
Django 2.0 breaks `if user.is_authenticated():` (utils.py, line 74), which needs to become `if user.is_authenticated:` -- this should also work from Django 1.10 onwards, see: https://docs.djangoproject.com/en/2.0/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods

bittner:
Would you mind to prepare a PR?

Anteru:
PR added.

Anteru:
Guess this can be closed then, given the PR has been merged. Hope you'll push out a release soon :)

bittner:
Thanks @Anteru!

PR #121, opened 2017-12-02
Use user.is_authenticated for Django 2.0 compatibility.

Anteru:
Fix for #120.

### DIFF
--- a/analytical/utils.py
+++ b/analytical/utils.py
@@ -49,6 +49,19 @@ def get_user_from_context(context):
     return None
 
 
+def get_user_is_authenticated(user):
+    """Check if the user is authenticated.
+
+    This is a compatibility function needed to support both Django 1.x and 2.x;
+    Django 2.x turns the function into a proper boolean so function calls will
+    fail.
+    """
+    if callable(user.is_authenticated):
+        return user.is_authenticated()
+    else:
+        return user.is_authenticated
+
+
 def get_identity(context, prefix=None, identity_func=None, user=None):
     """
     Get the identity of a logged in user from a template context.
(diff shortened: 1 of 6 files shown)


jcassee:
Thanks, for the PR, @Anteru, but django-ananytical still supports Django 1.8 (and thus Python 2). Could you update the PR 1) include Django 2.0 to the test set, and 2) fix backward-compatibility with 1.8?

jcassee:
LGTM, @bittner?

bittner:
Django 2.0 needs to be added to [.travis.yml](https://github.com/jcassee/django-analytical/blob/master/.travis.yml). (_env_ section, _matrix/exclude_ section) and [setup.py](https://github.com/jcassee/django-analytical/blob/master/setup.py#L81) (trove).

analytical/utils.py
```diff
@@ -71,7 +83,7 @@ def get_identity(context, prefix=None, identity_func=None, user=None):
         try:
             if user is None:
                 user = get_user_from_context(context)
-            if user.is_authenticated():
+            if get_user_is_authenticated(user):
                 if identity_func is not None:
```
bittner: I know this is not part of the changes of this PR, but ...

Simply `if identity_func:` would be more pythonic.

analytical/utils.py
```diff
@@ -49,6 +49,18 @@ def get_user_from_context(context):
     return None
 
 
+def get_user_is_authenticated(user):
+    """Check if the user is authenticated.
+
+    This is a compatibility function needed to support both Django 1.x and 2.x;
+    Django 2.x turns the function into a proper boolean so function calls will
+    fail."""
```
bittner: I would prefer to have the `"""` in a separate line, both for the start and the end of the docstring.

analytical/utils.py
jcassee: Are you sure? `is not None` is more explicit I think?

jcassee:
Thanks, Matthäus. I'll release a new version with this change shortly.

jcassee:
Released as 2.4.0.

PR #127, opened 2018-05-17
Fix the ordering of Google Analytics calls

vdboor:
Currently, django-analytical renders the google analytics like in the following order:

```
_gaq.push(['_setAccount', 'UA-########-##']);
_gaq.push(['_trackPageview']);
_gaq.push (['_gat._anonymizeIp']);
```

It's a bit strange though to call `_anonymizeIp` after the `_trackPageview` call already happened.
This patch changes the ordering, to follow:

```
_gaq.push(['_setAccount', 'UA-########-##']);
_gaq.push(['_gat._anonymizeIp']);
_gaq.push(['_trackPageview']);
```

This is what every blog recommends doing.

### DIFF
--- a/analytical/templatetags/google_analytics.py
+++ b/analytical/templatetags/google_analytics.py
@@ -32,7 +32,6 @@ SETUP_CODE = """
 
       var _gaq = _gaq || [];
       _gaq.push(['_setAccount', '%(property_id)s']);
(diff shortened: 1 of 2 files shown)


jcassee:
Fixed in deba4e0. Thanks @vdboor!

PR #128, opened 2018-07-27
docs: Piwik is now Matomo

sikmir:
See [Piwik is now Matomo – Announcement](https://matomo.org/blog/2018/01/piwik-is-now-matomo/)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,7 +65,7 @@ Currently Supported Services
 * `Olark`_ visitor chat
 * `Optimizely`_ A/B testing
 * `Performable`_ web analytics and landing pages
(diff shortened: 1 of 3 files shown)


bittner:
LGTM

bittner:
@jcassee Should we maybe add a deprecation warning and and a `matomo` copy of the `piwik` module to the source code?

bittner:
The build is failing, because [wheel](https://pypi.org/project/wheel/) has stopped supporting Python 3.3. We'll eventually have to drop Python 3.3 support.

bittner:
Thanks, @sikmir!

Issue #108, opened 2017-06-14
Any support planned for the newer Google analytics.js library?


qurm:
Hi team,

I see that the ga.js, has been replaced by a later library, and that is recommended for new implementations.  Would be great if you are able to support that also, perhaps in parallel to the ga.js for some time.
https://developers.google.com/analytics/devguides/collection/gajs/

Thanks for all your good work with djang-analytical,
Andy

jcassee:
Thanks for the heads-up! The JavaScript snippets are really just strings in Python sources. Do you think you could submit the changes yourself in the form of a pull request?

jcassee:
Merged #135

PR #135, opened 2018-09-24
Basic support of Google analytics.js, #108

mbourqui:
This PR adds the support of the the new [analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/) library from Google.

It provides a new `google_analytics_js` tag, which is included if the `GOOGLE_ANALYTICS_JS_PROPERTY_ID` is set. For backwards compatibility, and since it is intended to completely replace the legacy `ga.js` library, the settings are set using the same `GOOGLE_ANALYTICS_` prefix.


jcassee:
Thank you, @mbourqui! Sorry for the long wait.

PR #140, opened 2018-12-11
Add basic support of Google gtag.js

mbourqui:
This PR adds the support of the the new [gtag.js](https://developers.google.com/analytics/devguides/collection/gtagjs/) library from Google.

It provides a new `google_analytics_gtag` tag, which is included if the `GOOGLE_ANALYTICS_GTAG_PROPERTY_ID` is set. There are yet not settings supported.

### DIFF
--- a/analytical/templatetags/google_analytics_gtag.py
+++ b/analytical/templatetags/google_analytics_gtag.py
@@ -1,0 +1,65 @@
+"""
+Google Analytics template tags and filters, using the new analytics.js library.
+"""
+
+from __future__ import absolute_import
+
+import re
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import (
+    disable_html,
+    get_required_setting,
+    is_internal_ip,
+)
+
+PROPERTY_ID_RE = re.compile(r'^UA-\d+-\d+$')
+SETUP_CODE = """
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src="https://www.googletagmanager.com/gtag/js?id={property_id}"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+
+  gtag('config', '{property_id}');
+</script>
+"""
+
+register = Library()
+
+
+@register.tag
+def google_analytics_gtag(parser, token):
+    """
+    Google Analytics tracking template tag.
+
+    Renders Javascript code to track page visits.  You must supply
+    your website property ID (as a string) in the
+    ``GOOGLE_ANALYTICS_JS_PROPERTY_ID`` setting.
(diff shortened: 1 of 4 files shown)


analytical/templatetags/google_analytics_gtag.py
```diff
@@ -0,0 +1,64 @@
...
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import (
+    disable_html,
+    get_required_setting,
+    is_internal_ip,
+)
+
+PROPERTY_ID_RE = re.compile(r'^UA-\d+-\d+$')
```
bittner: If the ID should be `UA-XXXXXX-Y`, as you state below, then the regex should probably allow only a single number after the second dash, e.g. look like `r'^UA-\d+-\d$'`. Correct?

mbourqui: As per the [Google documentation](https://support.google.com/analytics/answer/7372977?ctx=glossary), the second number is the property number. There is no explicit constraint that it has to be less than 10. As such, the property number may be larger than 9, and restricting to a single digit may be too restrictive.

mbourqui:
I already use it on production for some projects, and so far I can confirm it is functioning.

analytical/tests/test_tag_google_analytics_gtag.py
```diff
@@ -0,0 +1,48 @@
...
+
+@override_settings(GOOGLE_ANALYTICS_GTAG_PROPERTY_ID='UA-123456-7')
+class GoogleAnalyticsTagTestCase(TagTestCase):
+    """
+    Tests for the ``google_analytics_js`` template tag.
+    """
+
+    def test_tag(self):
+        r = self.render_tag('google_analytics_gtag', 'google_analytics_gtag')
+        self.assertTrue("""<script async src="https://www.googletagmanager.com/gtag/js?id=UA-123456-7"></script>""" in r, r)
```
jcassee: This line (and [line 38](#diff-5faab0fb53bde9f058fe071a8cbeccdfR28)) is what flake8 complains about. The shortest form that I still find readable I came up with is:
```
        self.assertTrue('<script async src="https://www.googletagmanager.com/gtag/js?id=UA-123456-7">'
        	'</script>' in r, r)
```

Still 102 characters. Maybe add `# noqa: E501` here instead. @mbourqui @bittner what do you think?

mbourqui: Splitting it on three lines seems to do the trick, without breaking the html code line.

bittner:
Could you rebase your branch so the PR could be merged?

mbourqui:
> Could you rebase your branch so the PR could be merged?

Done!

bittner:
Thank you and sorry for the long wait! :rabbit: :egg: :rocket:

PR #146, opened 2019-04-06
Create Matomo module; leave Piwik to be deprecated

### LINKED ISSUE #132: Piwik module should be renamed to Matomo
As commented in PR #128, Piwik has been renamed to Matomo in the past. This should also reflect in the name of the related module.

Ideally, we should use a deprecation path, duplicating the modules while the deprecation warning is active.

sckarlin:
With the rebranding of Piwik to Matomo, this commit:
* copies the piwik module to matomo and rebrands
* notes that the piwik module is deprecated
* updates the javascript to the current Matomo version

Fixes #132

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (C) 2011-2016 Joost Cassee and others
+Copyright (C) 2011-2019 Joost Cassee and others
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (C) 2011-2019 Joost Cassee and others
+Copyright (C) 2011-2019 Joost Cassee and contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/analytical/templatetags/matomo.py
+++ b/analytical/templatetags/matomo.py
@@ -1,5 +1,5 @@
 """
-Piwik template tags and filters.
+Matomo template tags and filters.
 """
 
 from __future__ import absolute_import
@@ -14,8 +14,6 @@ from django.template import Library, Node, TemplateSyntaxError
 from analytical.utils import (is_internal_ip, disable_html,
                               get_required_setting, get_identity)
 
-import warnings
-warnings.warn('The Piwik module will be deprecated', DeprecationWarning)
 
 # domain name (characters separated by a dot), optional port, optional URI path, no slash
 DOMAINPATH_RE = re.compile(r'^(([^./?#@:]+\.)*[^./?#@:]+)+(:[0-9]+)?(/[^/?#@:]+)*$')
@@ -25,17 +23,17 @@ SITEID_RE = re.compile(r'^\d+$')
 
 TRACKING_CODE = """
 <script type="text/javascript">
-  var _paq = _paq || [];
+  var _paq = window._paq || [];
   %(variables)s
   %(commands)s
   _paq.push(['trackPageView']);
   _paq.push(['enableLinkTracking']);
   (function() {
     var u="//%(url)s/";
-    _paq.push(['setTrackerUrl', u+'piwik.php']);
+    _paq.push(['setTrackerUrl', u+'matomo.php']);
     _paq.push(['setSiteId', %(siteid)s]);
     var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
-    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
   })();
 </script>
 <noscript><p><img src="//%(url)s/piwik.php?idsite=%(siteid)s" style="border:0;" alt="" /></p></noscript>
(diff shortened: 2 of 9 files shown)


analytical/templatetags/piwik.py
```diff
@@ -14,6 +14,8 @@
 from analytical.utils import (is_internal_ip, disable_html,
                               get_required_setting, get_identity)
 
+import warnings
+warnings.warn('The Piwik module will be deprecated', DeprecationWarning)
```
jcassee: Maybe direct users to the Matomo module? (Also, nitpick, I think the Piwik module _is_ deprecated, not _will be_ deprecated.)

docs/services/piwik.rst
```diff
@@ -9,6 +9,13 @@ will always be yours, because you run your own analytics server.
 .. _Piwik: http://www.piwik.org/
 
 
+Deprecation Path
+================
+
+Note that Piwik is now known as Matomo.  New projects should use the
+Matomo integration.  The Piwik integration in django-analytical will
+eventually be deprecated.
```
jcassee: _Is_ deprecated and eventually removed?

docs/services/piwik.rst
```diff
@@ -1,5 +1,5 @@
 ==================================
-Matomo (formerly Piwik) -- open source web analytics
+Piwik (Deprecation Path) -- open source web analytics
```
jcassee: "(Deprecation Path)" -> "(deprecated)"?

analytical/templatetags/piwik.py
sckarlin: Clarified, thanks.

docs/services/piwik.rst
sckarlin: Clarified, thanks.

docs/services/piwik.rst
sckarlin: Clarified, thanks.

sckarlin:
Not sure what to do about the failing tests.  They seem unrelated to this PR.

jcassee:
Indeed, those errors are unrelated. I'm merging your changes. Thanks for the contribution!

sckarlin:
Super.  Thank you all for django-analytical.

bittner:
Yes, the issues are unrelated, but as the PRs are configured it may be best to fix them on a separate branch first and rebase this PR on the fixed `master` branch afterwards. I can try, if you guys want.

sckarlin:
@bittner : that would be great.

jcassee:
True, I'll fix master.

bittner:
@jcassee @sckarlin What I'm actually worried about in this PR is that we're unnecessarily duplicating code.

* Could make the Piwik implementation empty, making it just import everything from the new Matomo module, plus emitting a deprecation warining?
* The tests should stay for Piwik to make sure users running on the deprecated Piwik module are still served successfully.
* The content of the documentation chapter for Piwik should be replaced an explicit deprecation warning, and a link to the new Matomo chapter.

The point is, when we're removing the Piwik code we should have prepared everything. It should be obvious and easy to understand which parts to remove (and those parts should be really small -- to avoid human error!).

Does that sound okay?

sckarlin:
My thought is that the piwik module should stick around for a bit longer (6-12 months?) for those folks who may still be using a clone of the old Piwik code on there own servers which relies on the old names.

bittner:
Sure! It can stick around for even longer! It probably will. And that's why we should prepare the phase-out. That's what I say.

analytical/templatetags/matomo.py
```diff
@@ -0,0 +1,118 @@
...
+    (default) or ``'visit'``.  Index should be an integer and the
+    other parameters should be strings.
+    """
+    bits = token.split_contents()
+    if len(bits) > 1:
+        raise TemplateSyntaxError("'%s' takes no arguments" % bits[0])
+    return MatomoNode()
+
+
+class MatomoNode(Node):
```
bittner: 1. Make `PiwikNode` empty, make it inherit from this implementation (i.e. `PiwikNode(MatomoNode)`), and have it emit a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

analytical/templatetags/matomo.py
```diff
@@ -0,0 +1,118 @@
...
+DEFAULT_SCOPE = 'page'
+
+MatomoVar = namedtuple('MatomoVar', ('index', 'name', 'value', 'scope'))
+
+
+register = Library()
+
+
+@register.tag
+def matomo(parser, token):
```
bittner: 2. Make [the `piwik` tag](https://github.com/jazzband/django-analytical/blob/master/analytical/templatetags/piwik.py#L55) call this function, and make the `piwik` tag call emit a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

analytical/templatetags/piwik.py
```diff
@@ -14,6 +14,8 @@
 from analytical.utils import (is_internal_ip, disable_html,
                               get_required_setting, get_identity)
 
+import warnings
+warnings.warn('The Piwik module is deprecated; use the Matomo module.', DeprecationWarning)
```
bittner: Alright, that covers the suggested warning mentioned above. :heavy_check_mark:

bittner: The Piwik integration code here is irrelevant. It should simply inherit from Matomo.

sckarlin:
OK, so then both the Piwik and the Matomo modules need to co-exist with the Piwik module frozen as-is (using old interface names) and the Matomo module free to move forward with new features (such as proposed in #142).  Correct?

bittner:
No, the total opposite! I don't think it's worth conserving the old Piwik integration code. That should move forward just as Piwik/Matomo moves forward. If there's anything serious to implement in a backward-compatible fashion that should go into the new module.

The Piwik module should become just a hollow facade, a proxy to the new Matomo integration.

Remember: It's really all about renaming.

bittner:
Hmm, I believe this is the right way. If that proves to break the integration with an older Piwik instance developers can still stick to the current or an older django-analytical version, can't they?

Maybe I'm wrong. Let's discuss.

sckarlin:
This _seems_ a little more complicated than necessary to me.  Perhaps I'm not understanding.  If we were to have only one module named matomo going forward, are you suggesting we put conditionals in the python code so that this module works with Piwik servers, if needed?  For example, the javascript snippet used for the last Piwik-branded version has since diverged from the current Matomo-branded version.

I do agree that having a fair amount of duplicate code between the Piwik module and the Matomo module is not ideal; however, if we consider the Piwik module "frozen" then when the time comes to remove it completely, it'll be much easier.  By "frozen," I'm suggesting there would be no back-porting of new features for Matomo that theoretically could work with Piwik.

sckarlin:
Good point about using an older version of django-analytical for older Piwik support.

bittner:
Alright, let's leave the Piwik as it is. Sorry for the noise.

You can now rebase your branch onto the new `master`, and push the result to this PR. That should make all failing builds go away.

bittner:
You should also add yourself to the [AUTHORS.rst](https://github.com/jazzband/django-analytical/blob/master/AUTHORS.rst) file and update the date information in the [LICENSE.txt](https://github.com/jazzband/django-analytical/blob/master/LICENSE.txt) file.

sckarlin:
All done.  Thanks for your guidance.  Good discussion.

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright (C) 2011-2016 Joost Cassee and others
+Copyright (C) 2011-2019 Joost Cassee and others
```
bittner: Can you make this, "... and contributors"?

sckarlin: Done

sckarlin:
I performed an updated rebase.  Let me know if there are any issues.

bittner:
Thanks @sckarlin!

bittner:
Should we release a version 2.6.0 now, @jcassee?

jcassee:
> Should we release a version 2.6.0 now, @jcassee?

Absolutely. Will you or shall I?

bittner:
I think we both will have to be involved! :smirk: As it must go through a PR.

Can you create the PR? I'm interested whether push a tag to the master branch will work afterwards. :grimacing:

jcassee:
@bittner I think according to https://jazzband.co/about/releases we can just push a release commit?

bittner:
Try and we'll see what happens.

PR #147, opened 2019-04-08
Fix flake8 and readme tests

jcassee:
This PR fixes two problems:
* Remove the typing hints from `intercom.py` because they cause warnings in flake8. We cannot import the typing names because they don't exist in Python 2.
* Use the xenial distribution for the readme test. Django now requires Sqlite version 3.8.3 or higher, which is not available on trusty.

### DIFF
--- a/analytical/templatetags/intercom.py
+++ b/analytical/templatetags/intercom.py
@@ -8,8 +8,8 @@ import hashlib
 import hmac
 import json
 import sys
-import time
 import re
+import time
 
 from django.conf import settings
 from django.template import Library, Node, TemplateSyntaxError
@@ -29,7 +29,7 @@ TRACKING_CODE = """
 register = Library()
 
 
-def _timestamp(when):  # type: (datetime) -> float
+def _timestamp(when):
     """
     Python 2 compatibility for `datetime.timestamp()`.
     """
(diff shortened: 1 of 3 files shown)


bittner:
Take the readme change out! PR #148 fixes this properly! :smile:

bittner:
Be warned! I'll push the changes on top of this PR. :bomb: :smirk:

analytical/templatetags/intercom.py
```diff
@@ -8,8 +8,8 @@
 import hmac
 import json
 import sys
-import time
 import re
+import time
```
bittner: Is there a specific reason why you (or our editor) is changing the import order? Aren't they all standard library modules?

tox.ini
```diff
@@ -33,8 +33,10 @@ deps = flake8
 commands = flake8
 
 [testenv:readme]
-deps = readme_renderer
-commands = python setup.py check --restructuredtext --strict
+deps = twine
+commands =
+    {envpython} setup.py -q sdist bdist_wheel
+    twine check dist/*
```
bittner: I took this over from PR #148. This is the recommended replacement for the deprecated `check` call on setuptools.

tox.ini
```diff
@@ -47,4 +49,5 @@ DJANGO =
     2.1: django21
 
 [flake8]
+exclude = .cache,.git,.tox,build,dist
```
bittner: This is relevant if you run Flake8 locally.

bittner:
I'm merging.

analytical/templatetags/intercom.py
jcassee: Alphabetic order, I guess...

jcassee:
Thanks for taking this PR to the finishing line, @bittner

PR #150, opened 2019-04-09
Ensure Tox targets clean and docs are working

bittner:
* Ensure Tox target clean is working (execution order)
* Ensure Tox target docs is working (remove deprecated reference)
* Update packaging information (Django 2.2, license)
* Remove problematic install_requires from setup

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
     license=package.__license__,
     description=package.__doc__.strip(),
     long_description=read_file('README.rst'),
+    long_description_content_type='text/x-rst',
     author=package.__author__,
     author_email=package.__email__,
     packages=[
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -99,7 +96,4 @@ def read_file(fname):
     url='https://github.com/jazzband/django-analytical',
     download_url='https://github.com/jazzband/django-analytical/archive/master.zip',
     cmdclass=cmdclass,
-    install_requires=[
-        'Django>=1.7.0',
-    ],
```
jcassee: Why not update the dependency version?

bittner: Sometimes people install packages like django-analytical before Django. Only to find out later that "for some reason" a "wrong Django version" was installed. It's not worth that pain.

We should assume Python developers are adults who can read READMEs or the docs if something breaks.

PR #153, opened 2019-04-18
Make sure new Google Analytics anonimyze IP calls also happen before sending pageviews

vdboor:
In #127 I made sure that the old `ga.js` analytics code used the proper ordering of calls. It seems that the new `analytics.js` code introduced by #135 has the same issue.

Currently it renders:

```
ga('create', 'UA-25383750-9', 'auto', {});
 
ga('send', 'pageview');
ga('set', 'anonymizeIp', true);
```

The send method should happen as last method call however.

I checked the contents of the `{commands}` placeholder, and it can only consist of `ga('set', ...)` commands.

### DIFF
--- a/analytical/templatetags/google_analytics_js.py
+++ b/analytical/templatetags/google_analytics_js.py
@@ -30,14 +30,12 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 }})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
 ga('create', '{property_id}', 'auto', {create_fields});
-{display_features}
-ga('send', 'pageview');
-{commands}
+{commands}ga('send', 'pageview');
 </script>
 """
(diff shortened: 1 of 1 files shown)


bittner:
The Bandit check [is allowed to fail](https://github.com/jazzband/django-analytical/blob/master/.travis.yml#L18-L19), currently. I've opened an issue to have us reminded about fixing it.

vdboor:
@jcassee thanks for the quick response!
I've fixed the tests errors. It turned out that the script output included an extra blank line now. I've updated the `commands` construction so it always outputs one command per line, and all tests pass.

jcassee:
@bittner GitHub added an "Update branch" button to merge master into the feature branch. Do you have objections to using that instead of asking for a rebase?

bittner:
I don't really know. It's certainly a convenience feature. If GitHub is doing a merge commit, as the description says, there's little to nothing that can go wrong. I wouldn't be worried.

vdboor:
Thanks! :-D

PR #156, opened 2020-03-03
docs: Fix simple typo, betweeen -> between

timgates42:
There is a small typo in docs/services/matomo.rst, docs/services/piwik.rst.

Should read `between` rather than `betweeen`.


bittner:
Thanks! :+1:

PR #157, opened 2020-04-11
Prepare release v2.6.0

bittner:
This PR makes a few changes to prepare the release of a new version v2.6.0, which includes the new implementation for Google Analytics gtag.

## Changes

- Explain the three Google Analytics options
- Update change log
- Bump version


docs/services/google_analytics_gtag.rst
```diff
@@ -34,10 +37,12 @@ template.  Insert the tag at the bottom of the HTML head::
     {% load google_analytics_gtag %}
     <html>
     <head>
-    {% google_analytics_gtag %}
     ...
     </head>
+    <body>
     ...
+    {% google_analytics_gtag %}
+    </body>
```
bittner, line 45: Adding a `<script>` tag to the HTML `<head>` is illegal. So I guess the setup should be like this.

@mbourqui, @jcassee: Can you confirm?

docs/services/google_analytics_js.rst
```diff
@@ -35,9 +38,11 @@ template.  Insert the tag at the bottom of the HTML head::
     <html>
     <head>
     ...
-    {% google_analytics_js %}
     </head>
+    <body>
     ...
+    {% google_analytics_js %}
+    </body>
```
bittner, line 45: Same here: Adding a `<script>` tag to the HTML `<head>` is illegal. I guess the setup should be like above to be valid.

@mbourqui, @jcassee: Can you confirm?

docs/services/google_analytics_gtag.rst
mbourqui: According to the [google doc](https://developers.google.com/analytics/devguides/collection/gtagjs#install_the_global_site_tag), it must be in that `<head>` tag:

> immediately after the \<head\> tag

docs/services/google_analytics_js.rst
mbourqui: Similarly for analytics.js, according to the [google doc](https://developers.google.com/analytics/devguides/collection/analyticsjs#the_google_analytics_tag), it must be in that `<head>` tag:

> near the top of the \<head\> tag and before any other script or CSS tags

docs/services/google_analytics_gtag.rst
jcassee: > Adding a <script> tag to the HTML <head> is illegal.

Are you sure? I think the `<script>` tag is metadata, and [`<head>` can contain metadata](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head)?

bittner: The docs you quote clearly state:

> immediately _after_ the `<head>` tag

So, it has to go into the `<body>` tag, (ideally) as the very first data item.

mbourqui: It literally means the opening `<head>` tag, as one can read it on [this help page](https://support.google.com/analytics/answer/1008080?hl=en&ref_topic=1008079#basic). There is some more explicit usage example on [that other help page](https://support.google.com/displayvideo/answer/7562276).

docs/services/google_analytics_js.rst
bittner: That really scares me: Running this sample code through the W3 validator service tells it's valid HTML.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <script>
      alert("Valid HTML with JavaScript in the HEAD tag!");
    </script>
    <title>Foo</title>
  </head>
  <body>
  </body>
</html>
```

How could Google, as a consortium member, drag the standards so low? :astonished:

docs/services/google_analytics_gtag.rst
bittner: Hmm, probably not clearly enough: The seem to say, "after the (opening) `<head>` tag". The [W3 validator service](https://validator.w3.org/nu/#textarea) proves this is allowed. :man_facepalming: :bomb:

bittner:
Alright,

- We clarified that the GA setup instructions are correct (i.e. the script tags actually go into `<head>...</head>`, as weird as it seems).
- I updated the change log and bumped the version number (minor component).

I believe, we're ready to release the baby. @jcassee?

bittner:
On thing is missing: We're not testing against the latest supported Python+Django version combinations. This is probably worth being updated. It would also officially drop Python 2 support.

bittner:
This is now ready for merging, I guess. @jcassee, as soon as you accept the review we can merge.

PR #160, opened 2020-06-26
Add user_id to Google Analytics GTag

sean-wallace:
This commit adds user_id support for the gtag google analytics implementation.

### DIFF
--- a/analytical/templatetags/google_analytics_gtag.py
+++ b/analytical/templatetags/google_analytics_gtag.py
@@ -10,6 +10,7 @@ from django.template import Library, Node, TemplateSyntaxError
 
 from analytical.utils import (
     disable_html,
+    get_identity,
     get_required_setting,
     is_internal_ip,
 )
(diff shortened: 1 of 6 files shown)


analytical/templatetags/google_analytics_gtag.py
```diff
@@ -51,8 +55,18 @@ def __init__(self):
...
+        identity = get_identity(context)
+        if identity is not None:
+            other_fields['user_id'] = identity
+
+        extra = '\n'.join([
+            GTAG_SET_CODE.format(key=key, value=value) for key, value in other_fields.items()
+        ])
         html = SETUP_CODE.format(
             property_id=self.property_id,
+            extra=extra
```
bittner, line 69: A trailing comma might be nice on this line.

sean-wallace:
> I believe, you need to extend the module documentation, don't you?

I've added a little something.

> What is the `l` -> `tl` name changes all about? If those are meant to be "meaningful names" can you use _actual_ self-explanatory names, please?

These were the simplest changes that would let `flake8` pass -- it looks like some new rules have come in since the last time it ran on master. I undid it since I don't know what the single letter variables stand for.

analytical/templatetags/google_analytics_gtag.py
sean-wallace: Done, thanks.

bittner:
* `p` is the "position", I'd suggest to use `pos`.
* `l` is the "location", I'd suggest to use `loc`.

sean-wallace:
All done, thanks for the help!

bittner:
Thanks! :+1:

PR #161, opened 2020-06-28
remove unnecessary if statement

renantamashiro:
Remove unnecessary if statement and improve coverage

### DIFF
--- a/analytical/templatetags/uservoice.py
+++ b/analytical/templatetags/uservoice.py
@@ -56,7 +56,6 @@ class UserVoiceNode(Node):
         widget_key = context.get('uservoice_widget_key')
         if not widget_key:
             widget_key = self.default_widget_key
-        if not widget_key:
             return ''
         # default
         options = {}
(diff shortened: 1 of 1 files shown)


bittner:
I fear there is a misunderstanding:

Line 58 updates the value of `widget_key`, which is then re-evaluated. In theory, the second `if` could be moved into the first one, but other than that the code should be correct as it is.

Or do you see some other reasons for the need of your change?

analytical/templatetags/uservoice.py
```diff
@@ -56,8 +56,8 @@ def render(self, context):
         widget_key = context.get('uservoice_widget_key')
         if not widget_key:
             widget_key = self.default_widget_key
```
bittner, line 58: This could probably be written as:
```python
widget_key = context.get('uservoice_widget_key') or self.default_widget_key
```
Then there would only be a need for a single `if` subsequently, which might be less confusing.

renantamashiro:
Oh! Sorry, that's true! Maybe I've confused about how this if statement works.

Your last comment clarified it. Could I change with another commit?

PR #162, opened 2020-07-02
Only show piwik warning if using piwik

WhyNotHugo:
Code so far emits a warning to all users about about Piwik being
deprecated, even if Piwik is not being used.

Scope the warning so it only triggers for people to whom it's relevant.


bittner:
We should eventually remove the Piwiki integration as such, one day.

PR #170, opened 2020-11-25
Updated supported versions of Python and Django

smithdc1:
Removes now-unsupported Python versions from the test matrix, adds Python 3.9 and Django 3.1.


smithdc1:
@hramezani would you like to review this?

tox.ini
```diff
@@ -1,29 +1,30 @@
...
+    django31: Django>=3.1,<3.2
+
 commands =
     coverage run setup.py test
     coverage xml
 
 [gh-actions]
 python =
-    2.7: py27
     3.5: py35
```
bittner, line 23: `py35` should probably be removed if we want this to match the trove classifiers in `setup.py`.

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
```
bittner, line 11: `py35` should probably be removed if we want this to match the trove classifiers in `setup.py`.

bittner (changes requested):

setup.py
```diff
@@ -75,25 +72,24 @@ def read_file(fname):
...
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
-    ],
+        'Programming Language :: Python :: 3.9',
+       ],
     platforms=['any'],
+    python_requires='>=3.6',
+    install_requires=['django>=2.2'],
```
bittner, line 92: Adding this explicitly to the package is a debated topic. This has unwanted side-effects in some occasion (think: when django-analytical is installed before Django is installed).

I would suggest to leave this out.

hramezani:
This function is for Python2
https://github.com/jazzband/django-analytical/blob/4a07cb35e47343026182b4ee398693bbf625bc05/analytical/templatetags/intercom.py#L32-L38
Should be removed I think

Here is also we have old Python version related code:
https://github.com/jazzband/django-analytical/blob/2d8cbd25bd7d3fdd89cb56b13488ed6ad1e2da9b/analytical/tests/test_utils.py#L30-L38

setup.py
smithdc1: Yes... I've been told off as a releaing packages without this before :shrug:. I'll remove.

hramezani:
@smithdc1 I think we need to replace the usage of `_timestamp`. here is the search result of it https://github.com/jazzband/django-analytical/search?q=_timestamp

smithdc1:
Thanks. 

In addition I've still managed to miss a few blank lines.

smithdc1:
OK -- i think this is ready for another review. Let me know if you are happy and I'll squash.

Thank you.

hramezani:
Thanks @smithdc1, I also need to adopt GitHbu Actions https://github.com/jazzband/django-analytical/blob/master/.github/workflows/test.yml

smithdc1:
Hi Hasan, 

Sorry I don't quite understand your comment. Is there a project you have where you would like to move to gha? I've done it on a few of my projects and I am happy to help.

hramezani:
> Hi Hasan,
> 
> Sorry I don't quite understand your comment. Is there a project you have where you would like to move to gha? I've done it on a few of my projects and I am happy to help.

Sorry. My bad 🤦‍♂️. It was unrelated. I deleted my comment.
I will review it again soon

setup.py
```diff
@@ -75,24 +72,21 @@ def read_file(fname):
...
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
-    ],
+        'Programming Language :: Python :: 3.9',
+       ],
```
bittner, line 89: The indentation seems broken here.

tox.ini
```diff
@@ -1,29 +1,29 @@
 [tox]
 envlist =
     # Python/Django combinations that are officially supported
-    py{27,35,36,37}-django111
-    py{35,36,37,38}-django22
-    py{36,37,38}-django30
+    py{36,37,38,39}-django22
+    py{36,37,38,39}-django30
+    py{36,37,38,39}-django31
```
bittner, line 6: You could consolidate those 3 lines into a single one, e.g.
```suggestion
    py{36,37,38,39}-django{22,30,31}
```

bittner (approved):
I left two more comments that you could consider incorporating. Other than that, all fine.

hramezani (approved):
Thanks @smithdc1 . LGTM :+1:

bittner:
I just figured that in our GitHub Actions setup we don't separate the different Django versions from the Python versions that are being run.

- When a build job fails it will not be immediately clear what was the problem, whether it was about an entire Python version or a specific Python-Django combination.
- Visually, the local runs of Tox are split up, but the ones driven by GitHub Actions are grouped by Python version.

Frankly, this result irritates me, because we _do_ have [tox-gh-actions](https://github.com/jazzband/django-analytical/blob/master/.github/workflows/test.yml#L38) in the GHA setup. :thinking:  So, this actually _should_ work properly, already.

Examples of how the splitting-up may look like:
- [behave/behave-django](https://travis-ci.org/behave/behave-django) (Travis for Python & Django)
- [bittner/pyclean](https://github.com/bittner/pyclean/actions/runs/171161083) (GHA for Python & OSes)

bittner:
>  So, this actually _should_ work properly, already.

The [related documentation](https://github.com/ymyzk/tox-gh-actions#factor-conditional-settings-python-version) of `tox-gh-actions` is a bit irritating. I would assume, we only need to inject the Django version as a separate environment variable, similar to a setup with `tox-travis`, like so:

```ini
[gh-actions:env]
DJANGO =
    2.2: django22
    3.0: django30
    3.1: django31
```
and in [our GHA configuration](https://github.com/jazzband/django-analytical/blob/master/.github/workflows/test.yml#L10-L11) we need to populate them appropriately, e.g.

```yaml
      matrix:
        python-version:
        - 3.6
        - 3.7
        - 3.8
        - 3.9
        django-version:
        - 2.2
        - 3.0
        - 3.1
...
    - name: "${{ matrix.python-version }}-${{ matrix.django-version }}"
      run: tox
      env:
        DJANGO: ${{ matrix.django-version }}
```

smithdc1:
@bittner are you looking to address those comments here? They feel slightly out of scope for this PR to me. :thinking:

bittner:
> @bittner are you looking to address those comments here? They feel slightly out of scope for this PR to me. thinking

You're right. That warrants a PR on its own. It was just that I was shocked :zany_face: to find out this was missing.

bittner:
Now that #171 is merged there are conflicts with two files. Sorry about that. Can you resolve them and rebase your branch?

Then we're probably ready to merge.

smithdc1:
> You're right. That warrants a PR on its own. It was just that I was shocked  to find out this was missing.

No problem -- yes tox-gh-actions (is that what it is called?) does't quite split it out in the same way as Travis CI does it... Possibly a bit of feedback for @jezdez as he is working super hard on coverting the jazzband projects away from Travis CI. (In summary the issue here is the split by Django version is within the GitHub-Action run, rather than getting a full matrix of python/django combinatons)

> Sorry about that. Can you resolve them and rebase your branch?

Again no problem -- I proposed the change :+1:

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8']
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
jezdez, line 11: Don't remove quotes here. This is a good pattern to make sure that the next version of 3.10 won't be interpreted wrong in yml.

jezdez (changes requested):

smithdc1:
Ok. Re-based and added quotes. Hopefully we're there now with this.

jezdez:
@smithdc1 I've removed the requirement for 2.7 and 3.5 builds to pass before merge.

jezdez (approved):

bittner:
So, let's merge?

bittner:
Thanks @smithdc1, @hramezani, @jezdez! :trophy:

smithdc1:
No problem, it is a pleasure.

What do you think about a new release? I think it would be good to have a version that clearly supports Django 3.1?

bittner:
Sure!

I'd like to add the Lucky Orange integration (#172) first, though, if you don't mind. It's almost done. I have a last test that is failing...

Issue #117, opened 2017-10-21
Refactor setup.py (remove test setup, and the like)


skulegirl:
Installed from master today and ran into this error:

`django.core.exceptions.ImproperlyConfigured: Error loading either pysqlite2 or sqlite3 modules (tried in that order): No module named '_sqlite3'`

2.3.0 release doesn't seem to have this issue. Looks like commit 347df54502d4a921ff95aa0674712cdbd5c8fbf9 broke the setup.cfg.

Full text of error:
```
> pip install git+https://github.com/jcassee/django-analytical.git
Collecting git+https://github.com/jcassee/django-analytical.git
  Cloning https://github.com/jcassee/django-analytical.git to /tmp/pip-vtw00h9e-build
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "/webapps/bdw3_20171021121342Z/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 31, in <module>
        from pysqlite2 import dbapi2 as Database
    ModuleNotFoundError: No module named 'pysqlite2'
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/webapps/bdw3_20171021121342Z/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 33, in <module>
        from sqlite3 import dbapi2 as Database
      File "/usr/local/lib/python3.6/sqlite3/__init__.py", line 23, in <module>
        from sqlite3.dbapi2 import *
      File "/usr/local/lib/python3.6/sqlite3/dbapi2.py", line 27, in <module>
        from _sqlite3 import *
    ModuleNotFoundError: No module named '_sqlite3'
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-vtw00h9e-build/setup.py", line 49, in <module>
        django.setup()
      File "/webapps/bdw3_20171021121342Z/lib/python3.6/site-packages/django/__init__.py", line 27, in setup
        apps.populate(settings.INSTALLED_APPS)
      File "/webapps/bdw3_20171021121342Z/lib/python3.6/site-packages/django/apps/registry.py", line 108, in populate
        app_config.import_models()
      File "/webapps/bdw3_20171021121342Z/lib/python3.6/site-packages/django/apps/config.py", line 202, in import_models
        self.models_module = import_module(models_module_name)
[...]

PR #171, opened 2020-11-29
Use pytest as test runner

smithdc1:
* Used pytest to run tests within tox environments

* Removed test runner within setup.py

@bittner is this what you were thinking of to refactor the test runner out of setup.py?


setup.py
```diff
@@ -1,11 +1,9 @@
 import os
 
 try:
-    from setuptools import setup, Command
+    from setuptools import setup
 except ImportError:
-    from distutils.core import setup, Command
-
-os.environ['DJANGO_SETTINGS_MODULE'] = 'analytical.tests.settings'
+    from distutils.core import setup
```
bittner, line 6: I think this try-except is unneeded in modern Python 3.

setup.cfg
```diff
@@ -5,3 +5,6 @@ all_files  = 1
 
 [upload_sphinx]
 upload-dir = build/docs/html
+
+[tool:pytest]
+DJANGO_SETTINGS_MODULE = analytical.tests.settings
```
bittner, line 10: Newline missing at end of line.

tox.ini
```diff
@@ -10,11 +10,13 @@ envlist =
 description = Unit tests
 deps =
     coverage
+    pytest
```
bittner, line 13: You can probably omit that as a dependency when you specify `pytest-django`.

bittner (changes requested):
Not bad! That looks like a simple but effective change. Nice!  :+1: 

There are a few warnings in the `tox` runs, which however will be fix by PR #170, so we don't need to worry.

I made a few minor comments.

setup.py
```diff
@@ -17,42 +15,9 @@
...
-    user_options = []
-
-    def initialize_options(self):
-        pass
-
-    def finalize_options(self):
-        pass
-
-    def run(self):
-        from analytical.tests.utils import run_tests
```
smithdc1, line 31: I'll remove this function as well.

bittner:
Can you try to see if it's possible to also [remove the tests module](https://github.com/jazzband/django-analytical/blob/master/setup.py#L70-L71) from the packaging setup already?

```patch
--- a/setup.py
+++ b/setup.py
     packages=[
         'analytical',
         'analytical.templatetags',
-        'analytical.tests',
-        'analytical.tests.templatetags',
     ],
```

setup.py
```diff
@@ -17,42 +15,9 @@
...
-        "analytical.tests.settings"
-    )
-    django.setup()
-except ImportError:
-    print(
-        "Could not import django. "
-        "This is fine, unless you intend to run unit tests."
-    )
-
 import analytical as package  # noqa
```
bittner, line 21: Move this import statement up to its regular place, where all the others are, then you'll be able to remove the `# noqa` safely. Now this should be possible.

setup.py
```diff
@@ -17,42 +15,9 @@
     pass
```
bittner, line 15: The `cmdclass` variable and this entire block can also go away if you move the docs generation into `tox.ini` ([see example](https://github.com/behave/behave-django/blob/master/tox.ini#L45-L50)). :rocket:

setup.py
```diff
@@ -17,42 +15,9 @@
...
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
-
-try:
-    import django
-    os.environ.setdefault(
-        "DJANGO_SETTINGS_MODULE",
-        "analytical.tests.settings"
-    )
-    django.setup()
```
bittner, line 49: Great! Having this removed, we should now be able to declare #117 to be fixed! :rocket:

smithdc1:
Hi @bittner -- I've addressed your comments here. +7 / -72 seems like a nice win here. :rocket:

bittner (approved):
Sweet! I think we're ready to merge. :rocket: 

Good job! :+1:

PR #174, opened 2020-12-03
Split up test matrix into Python-Django combinations

bittner:
A test matrix shall give us a quick answer on what broke when some jobs fail. It's a bit like writing tests: "A test should ideally do just a single thing."

This change splits up the linear test jobs, mainly defined by their Python version in use, into all Python-Django combinations we test against. Basically, we let GitHub Actions do the same thing as Tox does, when we run the test suite locally:

```console
$ tox -l
py36-django22
py36-django30
py36-django31
py37-django22
py37-django30
py37-django31
py38-django22
py38-django30
py38-django31
py39-django22
py39-django30
py39-django31
...
```


bittner:
Looks like we lost the checks (linting & docs) this way. I'd suggest to add a `checks.yml` configuration to keeps this clean and obvious. This allows us to also rename them back to `flake8`, `bandit`, `readme` and `docs`, which is a bit nicer to read.

In general, I think we should only run the test suite for the PR and the main branch. Not for a simple `push` event, that runs the same tests twice and is unnecessary, provided we allow merging into the main branch only via a PR. Opinions? @jezdez?

bittner:
BTW, is RTD configured for django-analytical by the jazzband organization? Do the docs build automatically once we issue a release?

jezdez:
> Looks like we lost the checks (linting & docs) this way. I'd suggest to add a `checks.yml` configuration to keep this clean and obvious. This allows us also to rename them back to `flake8`, `bandit`, `readme` and `docs`, which is a bit nicer to read.

In theory I would welcome this change, but after doing a lot of those conversions I'm really worried how different the various Jazzband projects handle linting and am currently trialing https://pre-commit.ci/ to see if we can use that for the Linting/check pieces. Give me a few days to finish that trial please.

> In general, I think we should only run the test suite for the PR and the main branch. Not for a simple `push` event, that runs the same tests twice and is unnecessary, provided we allow merging into the main branch only via a PR. Opinions? @jezdez?

We want to test on push to PR branches as well, not just the main branch (here: "master") since any members can create feature branches in this repo. So it's more likely to happen than in other GitHub projects where PRs after more often created against a branch in a fork.

.github/workflows/test.yml
```diff
@@ -37,9 +51,10 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install --upgrade tox tox-gh-actions
 
-    - name: Tox tests
-      run: |
-        tox -v
+    - name: Unit tests for py${{ matrix.python-version }}-django${{ matrix.django-version }}
```
jezdez, line 54: 1. It's not relevant if those are  "unit tests" or any other type of tests
2. I want to prevent the GitHub config to have to be synced with the tox config as much as possible, so let's not use the env specifiers in descriptive text

```suggestion
    - name: Tox tests (Python ${{ matrix.python-version }}, Django ${{ matrix.django-version }})
```

tox.ini
```diff
@@ -24,30 +27,45 @@ python =
...
     twine check dist/*
 
+[testenv:clean]
+description = Clean up bytecode and build artifacts
+deps = pyclean
+commands =
+    pyclean -v {toxinidir}
+    rm -rf .tox/ build/ dist/ django_analytical.egg-info/ .coverage coverage.xml
+whitelist_externals =
+    rm
```
jezdez, line 67: Is this clean command just for development purposes?

jezdez (changes requested):

jezdez:
> BTW, is RTD configured for django-analytical by the jazzband organization?

Yes, that was done as part of the onboarding in #139.

> Do the docs build automatically once we issue a release?

Yes, it's configured to build on any push to master as the "latest" version, but Git tags are not automatically built. But Readthedocs has [new automation rules to do that](https://docs.readthedocs.io/en/stable/automation-rules.html). Please file a different ticket if you'd like to enable those (e.g. auto-enable versions when created).

.github/workflows/test.yml
bittner: I don't really care too much, but strictly speaking ... what are Tox tests at all? Tox is the test runner.

Granted, that's nit-picking.

tox.ini
bittner: Yes, this is only for local use.

bittner:
@jezdez Do you have time to look into this and take a decision?

Now that we're merging more PRs it irritates me again and again that we don't have checks, tests and the release process split up into different workflows. See [django-probes](https://github.com/painless-software/django-probes/tree/master/.github/workflows) for a real-world example ([user interface results](https://github.com/painless-software/django-probes/actions)).

bittner:
@jezdez Any hope that you may wrap your head around this topic? I think it makes sense to split the tests up into several jobs for the Django+Python combinations for all Django projects. That could be implemented generically across all projects, couldn't it?

bittner:
@jezdez There are 4 additional Python jobs (3.6, 3.7, 3.8, 3.9). They take this PR from completing the test suite. I don't see that they would come from this PR. Can you advise?

This PR would be ready to be merged, otherwise. :checkered_flag:

jezdez:
> @jezdez There are 4 additional Python jobs (3.6, 3.7, 3.8, 3.9). They take this PR from completing the test suite. I don't see that they would come from this PR. Can you advise?
> 
> This PR would be ready to be merged, otherwise. 🏁

They were from enforcing branch protection and the old required commit checks, which changed since the names are now different. I've removed them from the settings now but would still recommend requiring some checks to pass, would you agree?

jezdez (approved):
Let me know what checks to enforce, this looks good to me though!

bittner:
Thanks @jezdez! :trophy:

PR #175, opened 2020-12-05
Add Lucky Orange integration

### LINKED ISSUE #172: Lucky Orange integration
[Lucky Orange](https://www.luckyorange.com) is another website visitor surveillance service. They call themselves, the original all-in-one conversion optimization suite.

Basically, they're similar to Hotjar, but with a more aggressive business model (e.g. only 7 days free tier).

bittner:
Adds a new integration for a service similar to Hotjar.

Fixes #172

### DIFF
--- a/analytical/templatetags/luckyorange.py
+++ b/analytical/templatetags/luckyorange.py
@@ -1,0 +1,62 @@
+"""
+Lucky Orange template tags and filters.
+"""
+
+import re
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import get_required_setting, is_internal_ip, disable_html
+
+
+LUCKYORANGE_TRACKING_CODE = """\
+<script type='text/javascript'>
+window.__lo_site_id = %(LUCKYORANGE_SITE_ID)s;
+(function() {
+    var wa = document.createElement('script'); wa.type = 'text/javascript'; wa.async = true;
+    wa.src = 'https://d10lpsik1i8c69.cloudfront.net/w.js';
+    var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(wa, s);
+})();
+</script>
+"""
+
+
+register = Library()
+
+
+def _validate_no_args(token):
+    bits = token.split_contents()
+    if len(bits) > 1:
+        raise TemplateSyntaxError("'%s' takes no arguments" % bits[0])
+
+
+@register.tag
+def luckyorange(parser, token):
+    """
+    Lucky Orange template tag.
+    """
+    _validate_no_args(token)
+    return LuckyOrangeNode()
+
+
+class LuckyOrangeNode(Node):
+
+    def __init__(self):
+        self.site_id = get_required_setting(
+            'LUCKYORANGE_SITE_ID',
+            re.compile(r'^\d+$'),
+            "must be (a string containing) a number",
+        )
+
+    def render(self, context):
+        html = LUCKYORANGE_TRACKING_CODE % {'LUCKYORANGE_SITE_ID': self.site_id}
+        if is_internal_ip(context, 'LUCKYORANGE'):
+            return disable_html(html, 'Lucky Orange')
+        else:
+            return html
+
+
+def contribute_to_analytical(add_node):
+    # ensure properly configured
+    LuckyOrangeNode()
+    add_node('head_bottom', LuckyOrangeNode)
--- a/README.rst
+++ b/README.rst
@@ -62,6 +62,7 @@ Currently Supported Services
 * `Intercom`_ live chat and support
 * `KISSinsights`_ feedback surveys
 * `KISSmetrics`_ funnel analysis
+* `Lucky Orange`_ analytics and user feedback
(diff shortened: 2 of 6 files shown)


analytical/templatetags/luckyorange.py
```diff
@@ -0,0 +1,62 @@
...
+
+from analytical.utils import get_required_setting, is_internal_ip, disable_html
+
+
+LUCKYORANGE_TRACKING_CODE = """\
+<script type='text/javascript'>
+window.__lo_site_id = %(LUCKYORANGE_SITE_ID)s;
+(function() {
+    var wa = document.createElement('script'); wa.type = 'text/javascript'; wa.async = true;
+    wa.src = 'https://d10lpsik1i8c69.cloudfront.net/w.js';
```
PiDelport, line 17: Is this hostname dynamic, or safe to include statically?

I couldn't find any clear documentation from Lucky Orange about this.

bittner: I don't know either. I though, we'll find out. I would _assume_ this is a generated CloudFront URL that doesn't change. They would have hassle with their clients, otherwise.

I could, of course, open a support ticket with Lucky Orange to clarify this.

PiDelport: Is the code snippet from generic, shared documentation, or somewhere from within an individual account's interface on their service?

If it's from generic, shared documentation, it's probably fine to follow that and update the snippet as the documentation updates: it would be good to link to that from the code.

Otherwise, I'd be worried about the possibility of leaking account context or other account-specific information that might get included unintentionally in this?

It would be good to have official confirmation from Lucky Orange's support that the snippet and hostname is safe and secure to share between different account contexts.

bittner: You're right. You smelled it. It isn't from their documentation but rather from their settings interface in the user account. :+1: 

There are two reasons why I _assume_ that this is neither account-specific nor a secret we may accidentally leak:

1. The code snippet you are meant to include in your site is publicly available (sticking your nose in the generated website HTML code is sufficient).
1. Lucky Orange doesn't feel like a company to me that ia fully into Open Source and transparently publishing details; the lack of the code snippet in the docs might be a result of just that.

BTW, there is a [screenshot with a code snippet](https://help.luckyorange.com/article/123-where-can-i-find-my-tracking-code) in their public documentation, but it's hopelessly outdated. :zany_face: 

I'll to open a support ticket with them and ask for clarification.

PiDelport: Awesome, thank you.

On the balance, I can see that this is _probably_ okay, and that they probably just globally share that hostname (at least for now), but I think that django-analytical should err on the side of caution, and only merge this once we have explicit confirmation that the hostname is safe and secure to share globally, between all users of this project.

bittner: I understand your caution. Though, I would volunteer to sacrifice our Lucky Orange account for getting this merge in early. :smirk: 

If that code wouldn't work _at all_ for other `SITE_ID`s, i.e. other customers, that would be bad, though.

bittner: As it's Sunday and support won't answer immediately, I have taken a look into their [official integrations](https://help.luckyorange.com/article/95-installation-guides), e.g.

| Integration | URL used |
|----------|--------|
| [Drupal module](https://help.luckyorange.com/article/100-how-to-install-in-drupal)<br>(ZIP file in step 4) | Has `... ? "https://ssl" : "http://cdn") + ".luckyorange.com/w.js"` as in the aforementioned screenshot |
| [Joomla! extension](https://www.luckyorange.com/plugins/joomla/installation/installation/index.html)<br>(`joomla.zip` in Step 4) | Has `... ? "https://ssl" : "http://cdn") + ".luckyorange.com/w.js"` as in the aforementioned screenshot |
| [Wordpress plugin](https://wordpress.org/plugins/lucky-orange/)<br>(<kbd>Download</kbd> button) | Has `https://d10lpsik1i8c69.cloudfront.net/w.js` in the source code |

Hard to say for sure, but as Wordpress is the most widely used CMS worldwide it's probably safe to assume that will likely be the most up-to-date integration. It sports the same CloudFront URL we use, so they themselves don't worry.

smithdc1:
I proof read the docs and they seem good to me.

bittner:
Okay, I rebased the branch. So, we only need to decide whether we want to require explicit confirmation from Luky Orange support on the included CDN URL, before merging.

BTW, to implement this integration I simply copy+pasted & search+replaced the excellent Hotjar integration code, including the docs. And for the rebasing I made sure with a `diff` that everything is still super-aligned. So, I'm standing on the shoulders of others who are bigger, smarter and more hardworking than I am. :smile: – Thanks @PiDelport!

analytical/templatetags/luckyorange.py
PiDelport: Ah, cool, great catch! Thank you for doing that research. 💜️

I agree: if their official WordPress plugin uses that, we can assume it's safe and secure.

PiDelport (approved):

PR #181, opened 2020-12-08
Migrate tests to Pytest plain asserts

smithdc1:
When looking at the code when it is formatted with Black there are a number of cases where it would be a bit more readable if plain pytest asserts were used. 

I've done a few files as an example. Is this wanted? Let me know what you think.

### DIFF
--- a/tests/unit/test_tag_chartbeat.py
+++ b/tests/unit/test_tag_chartbeat.py
@@ -39,8 +39,8 @@ class ChartbeatTagTestCaseWithSites(TestCase):
         site = Site.objects.create(domain="test.com", name="test")
         with override_settings(SITE_ID=site.id):
             r = ChartbeatBottomNode().render(Context())
-            self.assertTrue(re.search('var _sf_async_config={.*"uid": "12345".*};', r), r)
-            self.assertTrue(re.search('var _sf_async_config={.*"domain": "test.com".*};', r), r)
+            assert re.search('var _sf_async_config={.*"uid": "12345".*};', r)
+            assert re.search('var _sf_async_config={.*"domain": "test.com".*};', r)
 
     @override_settings(CHARTBEAT_AUTO_DOMAIN=False)
     def test_auto_domain_false(self):
@@ -50,7 +50,7 @@ class ChartbeatTagTestCaseWithSites(TestCase):
         in _sf_async_config.
         """
         r = ChartbeatBottomNode().render(Context())
-        self.assertTrue('var _sf_async_config={"uid": "12345"};' in r, r)
+        assert 'var _sf_async_config={"uid": "12345"};' in r
 
 
 @override_settings(CHARTBEAT_USER_ID='12345')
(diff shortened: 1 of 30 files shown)


bittner (approved):
That's not only wanted, that's highly appreciated! We should migrate the entire test suite to Pytest style. This will make the test code much more concise.

And, yah, I'm glad we seem to agree [some Black formatting](https://github.com/jazzband/django-analytical/pull/180/files#diff-5174558889840525fe2a389e663c1b65976fff72d5acfa66f03e18336b63a56dR25-R27) is _reaaaaly_ awkward. :roll_eyes: :smile:

tests/unit/test_tag_chartbeat.py
```diff
@@ -95,6 +95,5 @@ def test_render_internal_ip(self):
         req.META['REMOTE_ADDR'] = '1.1.1.1'
         context = Context({'request': req})
         r = ChartbeatBottomNode().render(context)
-        self.assertTrue(r.startswith(
-                '<!-- Chartbeat disabled on internal IP address'), r)
-        self.assertTrue(r.endswith('-->'), r)
+        assert r.startswith('<!-- Chartbeat disabled on internal IP address')
+        assert r.endswith('-->')
```
bittner, line 99: Hey, you win! :trophy: That code is lovely to read. :love_you_gesture: See how it improved. Nice!

smithdc1:
Peter, I think this is done. I think I need to review again in a day or so with a fresh pair of eyes.

Are you still up for reviewing this? :laughing:

tests/unit/test_tag_facebook_pixel.py
```diff
@@ -41,46 +42,45 @@ class FacebookPixelTagTestCase(TagTestCase):
...
-        )
-        self.assertRaisesRegex(
-            TemplateSyntaxError,
-            r"^'facebook_pixel_body' takes no arguments$",
-            lambda: (Template('{% load facebook_pixel %}{% facebook_pixel_body "arg" %}')
-                     .render(Context({}))),
-        )
+        template = '{%% load facebook_pixel %%}{%% facebook_pixel_%s "arg" %%}'
+        with pytest.raises(TemplateSyntaxError, match="'facebook_pixel_head' takes no arguments"):
+            Template(template % "head").render(Context({}))
```
smithdc1, line 62: Oh my. Isn't this much better than a lambda :heart:

tests/unit/test_tag_crazy_egg.py
```diff
@@ -19,32 +20,33 @@ class CrazyEggTagTestCase(TagTestCase):
...
 
     def test_node(self):
         r = CrazyEggNode().render(Context())
-        self.assertTrue('/1234/5678.js' in r, r)
+        assert '/1234/5678.js' in r
 
     @override_settings(CRAZY_EGG_ACCOUNT_NUMBER=None)
     def test_no_account_number(self):
-        self.assertRaises(AnalyticalException, CrazyEggNode)
+        with pytest.raises(AnalyticalException):
```
smithdc1, line 31: I was less sure about these ones, but went for it as I think it helps with consistency. Would be good to come back and add the message to all of these but don't think I'm up for volunteering for that at the moment.

tests/unit/test_tag_facebook_pixel.py
bittner: Turn the `"'facebook_pixel_head' takes no arguments"` into a variable (e.g. `error_message`) then it will read, "match equals error message", which will be even nicer! :smiley: :heart_eyes:

tests/unit/test_tag_crazy_egg.py
bittner: camelCase is for Java devs!  :nerd_face:  We're into snake_case, baby. :snake: :sunglasses:

bittner:
> Peter, I think this is done. I think I need to review again in a day or so with a fresh pair of eyes.
> 
> Are you still up for reviewing this? laughing

Totally! It's half past one in the morning am I'm half-way through with reviewing your work of art. I should better go to bed... :sleeping:

smithdc1:
> Totally! It's half past one in the morning am I'm half-way through with reviewing your work of art. I should better go to bed...

Please don't burn out becuase of me as this isn't urgent. Take care of yourself.

tests/unit/test_tag_chartbeat.py
bittner: Because it reads better. It's more concise. It's less code that does the same. This is more maintainable. An investment into a sustainable future of the project.

bittner:
@smithdc1 We should finish this off an merge the changes. Do you think you have time to go over the code for a final round anytime soon?

smithdc1:
I can probably find time in the next week. What do you need me to do? I can review this again with a fresh pair of eyes?

(Happy new year btw! 🎉 - is it still to late to be saying this? 🤷‍♀️)

smithdc1:
@bittner --- just looking at this again and I've noticed that `Context` is misused in a fair number of places. For example:

https://github.com/jazzband/django-analytical/blob/896a19196f0ecca146e4cb8fb3b26e71e15642c8/tests/unit/test_tag_clickmap.py#L40-L41

Passing `Context` into render is wrong (although it works), instead this should be something like:
```python
r = ClickmapNode().render({'request': req})
```

Shall I tidy these up in this PR as well?

bittner:
> @bittner --- just looking at this again and I've noticed that `Context` is misused in a fair number of places. For example:
...
> Shall I tidy these up in this PR as well?

This PR is big enough. And that change might have side effects we don't manage to understand yet. Hence, I'd suggest to first merge this PR and then create a follow-up PR just for the fix of the (mis-)use of `Context`. What do you think?

smithdc1:
Happy to look at it as a follow up PR 🙂

bittner:
Shall we merge?

smithdc1:
We could do. If I spot anything else when I'm looking at `context` I can submit a follow up PR. 

I'm sure I'll have missed something but I think what we have here is a good enough change. 

Now is better than never? (And we're also not doing it 'right now!') 🤷‍♀️

bittner:
> Now is better than never? (And we're also not doing it 'right now!') woman_shrugging

Shoot! :grimacing: I'll be doing it "right now"! :fearful:

Issue #185, opened 2021-02-11
how to set enhanced eCommerce in google analytics


likecodeingloveproblems:
this page describe how to implement Google Analytics Enhanced eCommerce with gTag.
I don't know how to implement it?

bittner:
I'm not sure what your question is, but if you want to integrate GA in Django cleanly with [django-analytical](https://pypi.org/project/django-analytical/) and you're a total fresher, start here:

1. [Tutorial](https://django-analytical.readthedocs.io/en/latest/tutorial.html)
1. [Installation and configuration](https://django-analytical.readthedocs.io/en/latest/install.html)
1. [Google Analytics (gtag.js)](https://django-analytical.readthedocs.io/en/latest/services/google_analytics_gtag.html) (service details)

likecodeingloveproblems:
I know how to install and configure it and use it  in my project.
I know how to clone it.
but I don't know how to use cloned app in my existing project?

likecodeingloveproblems:
@bittner thanks for your response
this [page](https://developers.google.com/analytics/devguides/collection/gtagjs/enhanced-ecommerce) describe how to set enhanced eCommerce with gtag but i don't see any thing for this purpose in documentation and in code.
Is this feature available?

PR #186, opened 2021-02-12
adding event tag to gtag for enhanced ecommerce google analytics

likecodeingloveproblems:
as mentioned in #185  i modify google_analytics_gtag.py [https://github.com/jazzband/django-analytical/compare/master...likecodeingloveproblems:google-analytics4-tags?expand=1#diff-ec6f9c3ccabd11c1b836b9d8e52d1d4d41fe96f656220f7cdafe4e08846f9c01](url) to get set event data in gtag.
hopefully be useful

### DIFF
--- a/analytical/templatetags/google_analytics_gtag.py
+++ b/analytical/templatetags/google_analytics_gtag.py
@@ -27,7 +27,7 @@ SETUP_CODE = """
 """
 
 GTAG_SET_CODE = """gtag('set', {{'{key}': '{value}'}});"""
-
+GTAG_EVENT_CODE = """gtag('event', {key}, {value});"""
 register = Library()
 
 
@@ -61,9 +61,12 @@ class GoogleAnalyticsGTagNode(Node):
         if identity is not None:
             other_fields['user_id'] = identity
 
-        extra = '\n'.join([
+        commands = [
             GTAG_SET_CODE.format(key=key, value=value) for key, value in other_fields.items()
-        ])
+        ]
+        commands+=self._get_event_commands(context)
+
+        extra = '\n'.join(commands)
         html = SETUP_CODE.format(
             property_id=self.property_id,
             extra=extra,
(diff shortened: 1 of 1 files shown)


analytical/templatetags/google_analytics_gtag.py
```diff
@@ -72,6 +75,25 @@ def render(self, context):
...
 
+    def _get_event_commands(self, context):
+        values = (
+            context.get('google_analytics_event%s' % i) for i in range(1, 6)
+        )
+        params = [(i, v) for i, v in enumerate(values, 1) if v is not None]
+        commands = []
+        for _, var in params:
+            key = var[0]
+            value = var[1]
```
bittner, line 86: Maybe this could be written as a single line, like this?

```suggestion
            key, value = var
```

bittner (changes requested):
Please add ad least one test for the new functionality in the [related unit test module](https://github.com/jazzband/django-analytical/blob/master/tests/unit/test_tag_google_analytics_gtag.py).

From the test the expected outcome (the resulting JavaScript code) should be clear.

analytical/templatetags/google_analytics_gtag.py
```diff
@@ -72,6 +75,24 @@ def render(self, context):
...
+            context.get('google_analytics_event%s' % i) for i in range(1, 6)
+        )
+        params = [(i, v) for i, v in enumerate(values, 1) if v is not None]
+        commands = []
+        for _, var in params:
+            key, value = var
+            try:
+                dict(value)
+            except ValueError:
+                value = f"'{value}'"
```
bittner, line 89: This doesn't look like a valid pattern. The `try`-block shouldn't be use for testing on a data type. And what's code in the `except`-clause? I think this is an edge case of a "pythonic try-except" where the pattern doesn't fit.

It's not clear at first sight what we're trying to achieve. Maybe an `if instanceof(value, dict)` would be better here, because it's more explicit for the use case.

bittner (changes requested):
Please rebase your PR and add a test for the business logic you added. Thank you!

analytical/templatetags/google_analytics_gtag.py
bittner: Would you be interested to fix the code issue, so we can merge the changes?

Issue #188, opened 2021-02-22
how to set identity_func


likecodeingloveproblems:
In https://github.com/jazzband/django-analytical/blob/37009f2e08e78d41ddc1bdd1e940f195c3ae4352/analytical/utils.py#L64 identity_func is defined but in google_gtag there is not any way to set it?

SilverStrings024:
That's weird, I don't see it either. Does gtag not track users with identities? If they do then you could submit a PR to patch this.

bittner:
This is likely related to #189. Can you take a look?

PetrDlouhy:
Setting the identity is possible by context variable. Here is my attempt to better document how to do it: #219

Also the gtag provider is missing the prefix which is fixed in #217, for now only `analytical_identity` variable works.

PR #189, opened 2021-03-25
allow to set identity_func by ANALYTICAL_IDENTITY_FUNC settings

PetrDlouhy:
Google has in it's conditions for enabling UserID requirement, that prohibits sending personal data (such as e-mail address) to analytics.
I am using e-mail address as username, so using GTag would break the requirements for me.

This enables me to set user UUID for the UserID parameter.

### DIFF
--- a/analytical/utils.py
+++ b/analytical/utils.py
@@ -84,10 +84,12 @@ def get_identity(context, prefix=None, identity_func=None, user=None):
             if user is None:
                 user = get_user_from_context(context)
             if get_user_is_authenticated(user):
-                if identity_func is not None:
-                    return identity_func(user)
-                else:
+                identity_func = getattr(settings, 'ANALYTICAL_IDENTITY_FUNC',
+                                        identity_func)
+                if identity_func is None:
                     return user.get_username()
+                else:
+                    return identity_func(user)
         except (KeyError, AttributeError):
             pass
     return None
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -25,3 +25,17 @@ their default values.
     ``GOOGLE_ANALYTICS_INTERNAL_IPS`` to configure for Google Analytics.
 
     See :ref:`internal-ips`.
+
+.. data:: ANALYTICAL_IDENTITY_FUNC
+
+    Default: Identity function dependent on provider
+
+    A function that returns the identity of the given user. This overrides the
+    default settings of different providers.
+
+    E.g. Google has in its conditions for enabling UserID the requirement, that prohibits
+    sending personal data (such as an e-mail address) to analytics.
+    If e-mail address is used as username, using GTag would break the requirements.
+
+    In such case add uuid field to the user and set ```ANALYTICAL_IDENTITY_FUNC``` to
+    ```lambda user: user.uuid```
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -34,6 +34,7 @@ class SettingDeletedTestCase(TestCase):
 
 class MyUser(AbstractBaseUser):
     identity = models.CharField(max_length=50)
+    name = models.CharField(max_length=50)
     USERNAME_FIELD = 'identity'
 
     class Meta:
@@ -45,6 +46,11 @@ class GetIdentityTestCase(TestCase):
     def test_custom_username_field(self):
         get_id = get_identity(Context({}), user=MyUser(identity='fake_id'))
         assert get_id == 'fake_id'
+
+    @override_settings(ANALYTICAL_IDENTITY_FUNC=lambda user: f"{user.name} ({user.identity})")
+    def test_custom_username_field_identity_func(self):
+        get_id = get_identity(Context({}), user=MyUser(identity='fake_id', name='fake_name'))
+        assert get_id == 'fake_name (fake_id)'
 
 
 @override_settings(ANALYTICAL_DOMAIN="example.org")


bittner (changes requested):
Can you extend the tests to match your changes, please? And also mention `ANALYTICAL_IDENTIFY_FUNC` in the docs.

Thank you for your understanding.

analytical/utils.py
```diff
@@ -84,6 +84,8 @@ def get_identity(context, prefix=None, identity_func=None, user=None):
             if user is None:
                 user = get_user_from_context(context)
             if get_user_is_authenticated(user):
+                identity_func = getattr(settings, 'ANALYTICAL_IDENTIFY_FUNC',
+                                        identity_func)
                 if identity_func is not None:
                     return identity_func(user)
                 else:
```
bittner, line 91: I know this is not part of the PR, but if you wouldn't mind: While you're here touching the code, I'd like to revert the `if`-clause to reduce cognitive overhead, i.e.

```python
                if identity_func is None:
                    return user.get_username()
                else:
                    return identity_func(user)
```
Feel free to ignore this request.

PetrDlouhy:
@bittner I will look at it.

PetrDlouhy:
@bittner I added simple test, some docs and also renamed the setting from ANALITICAL_IDENTIFY_FUNC to ANALYTICAL_IDENTITY_FUNC.
Please approve the workflows.

docs/settings.rst
```diff
@@ -25,3 +25,17 @@ their default values.
     ``GOOGLE_ANALYTICS_INTERNAL_IPS`` to configure for Google Analytics.
 
     See :ref:`internal-ips`.
+
+.. data:: ANALYTICAL_IDENTITY_FUNC
+
+    Default: Identity function dependend on provider
```
bittner, line 31:
```suggestion
    Default: Identity function dependent on provider
```

docs/settings.rst
```diff
@@ -25,3 +25,17 @@ their default values.
...
+
+    A function that returns the identity of the given user. This overrides the
+    default settings of diferent providers.
```
bittner, line 34:
```suggestion
    default settings of different providers.
```

docs/settings.rst
```diff
@@ -25,3 +25,17 @@ their default values.
...
+
+    E.g. Google has in it's conditions for enabling UserID requirement, that prohibits
```
bittner, line 36:
```suggestion
    E.g. Google has in its conditions for enabling UserID the requirement, that prohibits
```

docs/settings.rst
```diff
@@ -25,3 +25,17 @@ their default values.
...
+    sending personal data (such as e-mail address) to analytics.
```
bittner, line 37:
```suggestion
    sending personal data (such as an e-mail address) to analytics.
```

bittner (changes requested):
IIUC, the problem you're trying to solve relates to hard requirements of some of Analytical's analytics providers. Shouldn't this be solved in the respective modules instead of requiring our users to code an identity function?

PetrDlouhy:
@bittner I have fixed the typos.

I think, that possibility to manually override the identity function allows flexibility in corner cases. E.g. I want to use my UUID field to keep consistency of my data even if the default behavior of `gtag` provider is fixed.

I don't know other providers, but the `django-analytical` might use hash from the username as identity function for `google_analytics_gtag` provider. But that would interrupt consistency of the collected data for users that are updating this app. So I am not sure how to update this correctly (Make a new provider and mark the current obsolete?).

It is also truth, that one identity function might not fit for all providers, if more of them are used. So maybe we might better use settings like:
```python
ANALYTICAL_IDENTITY_FUNC_OVERRIDES = {
    'google_analytics_gtag': lambda user: user.uuid,
}
```
and
```python
ANALYTICAL_IDENTITY_FUNC_DEFAULT = lambda user: user.uuid
```
which would override only the default `user.get_username()` function.

bittner (changes requested):
I agree that making the identity func dependent on the provider would make sense. I would keep it simple, though.

```python
ANALYTICAL_IDENTITY_FUNC = {
    'google_analytics_gtag': lambda user: user.uuid,
}
```

or maybe even

```python
ANALYTICAL_AUTO_IDENTIFY = {
    'google_analytics_gtag': lambda user: user.uuid,
    'matomo': True,
}
```

Though I just noticed something else: The feature you attempt to contribute already exists, doesn't it? Instead of a (lamda) function, though, you simply add an `analytical_identity` value (or the service specific value) to the template context. – Isn't that good enough?

docs/settings.rst
```diff
@@ -25,3 +25,17 @@ their default values.
...
+.. data:: ANALYTICAL_IDENTITY_FUNC
+
+    Default: Identity function dependent on provider
+
+    A function that returns the identity of the given user. This overrides the
+    default settings of different providers.
+
+    E.g. Google has in its conditions for enabling UserID the requirement, that prohibits
+    sending personal data (such as an e-mail address) to analytics.
+    If e-mail address is used as username, using GTag would break the requirements.
```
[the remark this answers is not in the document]
PetrDlouhy: I meant transition from `google_analytics` to `google_analytics_gtag` with user identity which I underwent in my project.

The requirements are their terms of use, which they write very suggestively: https://developers.google.com/analytics/solutions/crm-integration#user_id
And also I think I had to agree to that when switching on the User ID feature in Google Analytics.

analytical/utils.py
```diff
@@ -79,15 +79,17 @@ def get_identity(context, prefix=None, identity_func=None, user=None):
         return context['analytical_identity']
     except KeyError:
         pass
-    if getattr(settings, 'ANALYTICAL_AUTO_IDENTIFY', True):
+    if getattr(settings, 'ANALYTICAL_AUTO_IDENTITY', True):
```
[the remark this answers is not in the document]
bittner: It's in [the documentation](https://django-analytical.readthedocs.io/en/latest/features.html#identifying-visitors) like that.

PetrDlouhy:
You are right. The `analytical_identity` in context works. Maybe we should document it better (it is described under few providers, but not in general docs nor for `google_analytics_gtag` provider.

Also some providers (like `google_analytics_gtag`) doesn't pass the `prefix` parameter, so it is not possible to use provider specific value.

analytical/utils.py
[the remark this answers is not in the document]
PetrDlouhy: Yeah, sorry for that. I fixed it even thou we might not use this PR.

bittner:
> Maybe we should document it better (it is described under few providers, but not in general docs nor for `google_analytics_gtag` provider.

Great! It would be awesome if you could provide the missing pieces to the documentation. A working code sample would be fantastic.

> Also some providers (like `google_analytics_gtag`) doesn't pass the `prefix` parameter, so it is not possible to use provider specific value.

That was already reported (#188, #197). Can you spot how to fix this problem?

_P.S.: Too sad for the nice test you added just recently!_ :cry:

PR #193, opened 2021-07-12
Fix #141 and add custom event trigger binding support

### LINKED ISSUE #141: Asking for consent - Piwik
Hey guys,

I just realized that there is no way to use the "Ask for consent" feature of Matomo (Piwik). This is relevant because of the new GDPR in europe. I'm about to create a pull request for this one, but I want to hear your opinion on my implementation idea.

I want to introduce a `PIWIK_ASK_FOR_CONSENT` variable, like `PIWIK_DISABLE_COOKIES`.
To make it easy for the user of the consent functionalty to integrate the question for consent, I wrote event listeners which are triggered on click and are attached to DOM elements with the following id's:

`piwik_deny_consent`
`piwik_give_consent`

What do you think? Do see any problems with this kind of implementation?

SilverStrings024:
This should work to make Matomo respect consent.
I created two new utility functions, one builds paq commands, the other builds javascript to bind an event listener to elements with a certain class name. That will allow users to make their own custom template tags then use `analytical.utils.get_event_bind_js()` to bind events to their elements.

### DIFF
--- a/analytical/templatetags/matomo.py
+++ b/analytical/templatetags/matomo.py
@@ -7,15 +7,12 @@ from collections import namedtuple
 from itertools import chain
 
 from django.conf import settings
-from django.template import Library, Node, TemplateSyntaxError
-
+from django.template import Library, Node, TemplateSyntaxError, Template
 from analytical.utils import (
     disable_html,
     get_identity,
     get_required_setting,
-    is_internal_ip,
-    build_paq_cmd,
-    get_event_bind_js
+    is_internal_ip
 )
 
 # domain name (characters separated by a dot), optional port, optional URI path, no slash
@@ -97,19 +94,6 @@ class MatomoNode(Node):
         if getattr(settings, 'MATOMO_DISABLE_COOKIES', False):
             commands.append(DISABLE_COOKIES_CODE)
 
-        if getattr(settings, "MATOMO_REQUIRE_CONSENT", False):
-            grant_class_name = settings.GRANT_CONSENT_TAG_CLASSNAME
-            revoke_class_name = settings.REVOKE_CONSENT_CLASSNAME
-            commands.append(build_paq_cmd('requireConsent'))
-            commands.append(get_event_bind_js(
-                class_name=grant_class_name,
-                matomo_event="rememberConsentGiven",
-            ))
-            commands.append(get_event_bind_js(
-                class_name=revoke_class_name,
-                matomo_event="forgetConsentGiven",
-            ))
-
         userid = get_identity(context, 'matomo')
         if userid is not None:
             variables_code = chain(variables_code, (
@@ -122,8 +106,14 @@ class MatomoNode(Node):
             'variables': '\n  '.join(variables_code),
             'commands': '\n  '.join(commands)
         }
+        # Force the consent script to render so we can inject it into the template
+        consent_script = Template("{{consent_script}}").render(context)
+        if len(consent_script) > 1:
+            html += consent_script
+
         if is_internal_ip(context, 'MATOMO'):
             html = disable_html(html, 'Matomo')
+
         return html
 
 
--- a/analytical/templatetags/matomo.py
+++ b/analytical/templatetags/matomo.py
@@ -14,6 +14,8 @@ from analytical.utils import (
     get_identity,
     get_required_setting,
     is_internal_ip,
+    build_paq_cmd,
+    get_event_bind_js
 )
 
 # domain name (characters separated by a dot), optional port, optional URI path, no slash
@@ -48,9 +50,7 @@ DEFAULT_SCOPE = 'page'
 
 MatomoVar = namedtuple('MatomoVar', ('index', 'name', 'value', 'scope'))
 
-
 register = Library()
-
 
 @register.tag
 def matomo(parser, token):
@@ -96,6 +96,19 @@ class MatomoNode(Node):
         commands = []
         if getattr(settings, 'MATOMO_DISABLE_COOKIES', False):
             commands.append(DISABLE_COOKIES_CODE)
+
+        if getattr(settings, "MATOMO_REQUIRE_CONSENT", False):
+            grant_class_name = settings.GRANT_CONSENT_TAG_CLASSNAME
+            revoke_class_name = settings.REVOKE_CONSENT_CLASSNAME
+            commands.append(build_paq_cmd('requireConsent'))
+            commands.append(get_event_bind_js(
+                class_name=grant_class_name,
+                matomo_event="rememberConsentGiven",
+            ))
+            commands.append(get_event_bind_js(
+                class_name=revoke_class_name,
+                matomo_event="forgetConsentGiven",
+            ))
 
         userid = get_identity(context, 'matomo')
         if userid is not None:
--- a/analytical/utils.py
+++ b/analytical/utils.py
@@ -1,7 +1,7 @@
 """
 Utility function for django-analytical.
 """
-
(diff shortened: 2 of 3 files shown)


SilverStrings024:
Just so anyone seeing #141 can see this PR I'm tagging the issue

SilverStrings024:
@bittner would you mind looking over this PR? It's pretty simple but can expose some extra options for the users of the package

analytical/utils.py
```diff
@@ -163,3 +163,56 @@ class AnalyticalException(Exception):
...
+            used when building commands to be used on manual event trigger.
+        
+    :Returns:
+        A complete '_paq.push([])' command in string form
+    """
+    paq = "_paq.push(['%s', " % (cmd)
+    if len(args) > 0:
+        for arg_idx in range(len(args)):
+            if arg_idx == len(args)-1:
+                paq += "'%s'])" % (args[arg_idx])
```
SilverStrings024, line 181: Add a try/except to attempt and convert the value to a bool or int, if it's of either type then don't add the quotation marks

analytical/templatetags/matomo.py
```diff
@@ -97,6 +97,19 @@ def render(self, context):
         if getattr(settings, 'MATOMO_DISABLE_COOKIES', False):
             commands.append(DISABLE_COOKIES_CODE)
 
+        if getattr(settings, "MATOMO_REQUIRE_CONSENT", False):
```
SilverStrings024, line 100: Add something to check if registered users are to be tracked no matter what (since they normally have to accept TOS and such). Someone may have another opinion on the best option but I'm going to require a setting called `MATOMO_TRACK_REGISTERED` which, if true (and the user is logged in), will **not** push the requireConsent command to `_paq`

analytical/templatetags/matomo.py
```diff
@@ -97,6 +97,19 @@ def render(self, context):
...
+            grant_class_name = settings.GRANT_CONSENT_TAG_CLASSNAME
```
SilverStrings024, line 101: I'm not sure but this **might** cause issues for people using things like tailwindcss. Unless `document.getElementByClassName` can get elements despite the class attribute having multiple (space separated) values. This would need tested with that kind of set up

SilverStrings024 (commented):
Accidentally started a review...but this is kind of important considering it can possibly break the code

analytical/templatetags/matomo.py
SilverStrings024: **for anyone who sees this**: What do you think about making a very simple context provider that checks if the user is logged in, if so don't render the consent code if  `settings.ALWAYS_TRACK_REGISTERED` is `True`. Otherwise if the user is not logged in or that setting is set to False (or not set) pass the script to the template.
Then in mamoto.py in the template tag, make it render something to actually inject the script by adding something like this to the end of whatever scripts are returned. `{{consent_script}}`
If we ALWAYS have that part at the end it should always render correctly.

This is the only way I could come up with

SilverStrings024:
@jcassee @bittner I was looking into adding the ability to conditionally render the consent code depending on if the user is logged in or not and the only way I can see to do something like this is if we have direct access to the user object or the request object.
Considering how Django is structured, the tag and the context we need are in different scopes so one can't access the other.

However, I was looking into inclusion tags and I noticed that it's exactly what this package does. So would it be okay to change the way the Matomo tag works to be an inclusion tag as opposed to a normal tag? This way I can get the context that was provided to the template which means I'll either have direct access to the request or the request provider I wrote to do the conditional rendering.
So in a nutshell, this package returns html but is a normal, completely isolated tag. But if we change it to be an inclusion tag that grabs the context from `context_providers.request` we can make this package very versatile. Especially considering we're doing nothing but adding javascript to templates.

If you're okay with this, I'd like to work on the Matomo tag conversion. I already have the context provider, the `_paq.push` command builder, a function to build a script that binds whatever event you want to whatever element you provide the class name for (I can and will expand this to support more options than just the class name) so doing this shouldn't take very long at all.
Once I get the green light, I'll start working on it and, the sooner the better because a portion of my current project relies on this project.

SilverStrings024:
@bittner This seems ready, would you mind looking it over?

bittner:
- [ ]  AFAICS, you're making unrelated changes to `analytical/utils.py`. Would you mind moving that to a separate PR?
- [ ]  Tests are missing for your changes. Please supply them with your feature enhancement.
- [ ]  Please squash your changes. If the change set is small enough the PR should have only one or two commits (e.g. one for the tests, one for the feature enhancement).

SilverStrings024:
>     * [ ]   AFAICS, you're making unrelated changes to `analytical/utils.py`. Would you mind moving that to a separate PR?
> 
>     * [ ]   Tests are missing for your changes. Please supply them with your feature enhancement.
> 
>     * [ ]   Please squash your changes. If the change set is small enough the PR should have only one or two commits (e.g. one for the tests, one for the feature enhancement).

I guess I can see that but I figured dynamically creating the javascript for event binding could let people trigger custom events as well as keep the files *relatively* free of chunks of hard coded javascript so, it just made sense to add them in.

It may take me a minute to figure out how to properly write tests for this. I'll dig in on my next break from my current project.

Squashing the changes means using `git rebase` right? I've never done this before so, sorry for the inexperience.
So, from what I understand and, forgive me if I'm misunderstanding, I need to rebase all of the changes out, take out the functions I added in analytics/utils.py  then just hardcode the javascript into the context provider? Then on a separate PR introduce the two utility functions again? Wouldn't that be kind of redundant considering it would go from dynamic to harcoded then back and we'd have to change the files back to their original state?
Again, I apologize for any misunderstanding if there is one.

analytical/utils.py
```diff
@@ -163,3 +163,90 @@ class AnalyticalException(Exception):
...
+
+        return arg
+
+    paq = "_paq.push(['%s'" % (cmd)
+    if len(args) > 0:
+        paq += ", "
+        for arg_idx in range(len(args)):
+            current_arg = __to_js_arg(args[arg_idx])
+            no_quotes = type(current_arg) in [bool, int, dict, list]
+            if arg_idx == len(args)-1:
```
bittner, line 208: This is – and the `else` black below – is not an elegant solution. Why don't you simply always add the `", "` and replace it by `"]);"` after the end of the loop? e.g.

```python
   paq = paq[:-3)] + "]);"
   return paq
```

analytical/utils.py
```diff
@@ -163,3 +163,90 @@ class AnalyticalException(Exception):
...
+                    segment = "'%s', " % (current_arg)
+            paq += segment
+    else:
+        paq += "]);"        
+    return paq
+
+def get_event_bind_js(
+    class_name, matomo_event,
+    matomo_args=[], js_event="onclick",
+    ):
```
bittner, line 226:
```suggestion
def get_event_bind_js(
    class_name, matomo_event, matomo_args=[], js_event="onclick"
):
```

bittner (changes requested):
I left a few style comments.

More than that, there are no tests for your code. Please add some!

bittner:
> It may take me a minute to figure out how to properly write tests for this.

Take a look at the tests that are there. The idea is simple:

1. Your function generates output for some specific input. Hence, in your test code, call your function with some arguments.
2. After that, in the line afterwards (in the same test function) verify the result using `assert`.

That's it.

If there are various possible outcomes for different input, add more of those calls with different input, so that all of your `if`-`else` clauses are being traveled through. – Our test suite if full of examples. Take a look!

PR #194, opened 2021-07-16
Added Heap integration

garrettc:
This adds [heap](https://heap.io/) integration (also known as Heap Analytics).

It's pretty much a copy and paste of the Clickmap integration with some updated documentation. 

I've confirmed it works in my Heap account (dashboard screenshot below):

Heap dashboard screenshot


analytical/templatetags/heap.py
```diff
@@ -0,0 +1,54 @@
...
+
+from django.template import Library, Node, TemplateSyntaxError
+
+from analytical.utils import is_internal_ip, disable_html, get_required_setting
+
+
+HEAP_TRACKER_ID_RE = re.compile(r'^\d+$')
+TRACKING_CODE = """
+    <script type="text/javascript">
+  window.heap=window.heap||[],heap.load=function(e,t){window.heap.appid=e,window.heap.config=t=t||{};var r=document.createElement("script");r.type="text/javascript",r.async=!0,r.src="https://cdn.heapanalytics.com/js/heap-"+e+".js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(r,a);for(var n=function(e){return function(){heap.push([e].concat(Array.prototype.slice.call(arguments,0)))}},p=["addEventProperties","addUserProperties","clearEventProperties","identify","resetIdentity","removeEventProperty","setEventProperties","track","unsetEventProperty"],o=0;o<p.length;o++)heap[p[o]]=n(p[o])};
```
bittner, line 15: Flake8 complains that this line is too long. Can you split this up somehow w/o breaking the code?

analytical/templatetags/heap.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Heap tracker template tag.
+
+    Renders Javascript code to track page visits.  You must supply
+    your heap tracker ID (as a string) in the ``HEAP_TRACKER_ID``
+    setting.
+    """
+    bits = token.split_contents()
+    if len(bits) > 1:
+        raise TemplateSyntaxError("'%s' takes no arguments" % bits[0])
```
bittner, line 35: Can we add a test that verifies that this error is raised?

bittner (changes requested):
Flake8 fails (interestingly with Python 3.7 only) and isort complains about the sort order of the imports. Can you fix this?

If you also add another or two tests to please the coverage then you're my hero! :trophy:

analytical/templatetags/heap.py
garrettc: I've taken the quick way out and `#noqa`d the line. It's the Heap supplied code and I didn't fancy trying to unwrap that and be confident it wasn't going to break something at the Heap end. Hope that's okay?

analytical/templatetags/heap.py
garrettc: Added a test for this.

garrettc:
Thanks for the comments.

As well as the specific changes above I've also run `isort` and `flake8` over the files and fixed up a few other little whitespace errors.

bittner (approved):
Nice, thank you! Looks good to me.

Would you mind squashing the commits into a single one?

<s>Not sure why the test are not being triggered, hmmm.</s> :thinking: – I had to trigger them as you're a first-time contributor around here... :nerd_face:

PR #195, opened 2021-09-27
Updated supported Django versions

smithdc1:
Django 3.0 is now unsupported. Django 4.0 is still in alpha, release date is January 2022.

Source: https://www.djangoproject.com/download/


tests/testproject/settings.py
```diff
@@ -29,3 +29,5 @@
         'APP_DIRS': True,
     },
 ]
+
+USE_TZ = False
```
smithdc1, line 33: This should probably be `True` but would need more work on the tests. As that's not particularly important to this package I'm not sure it's a concern?

tests/unit/test_utils.py
```diff
@@ -37,7 +37,7 @@ class MyUser(AbstractBaseUser):
     USERNAME_FIELD = 'identity'
 
     class Meta:
-        abstract = True
+        abstract = False
```
smithdc1, line 40: I'm not sure about this change, again maybe it is ok here but it seems like a breaking change in Django and I can't see it in the release notes.

tests/testproject/settings.py
smithdc1: It was a 3.2 change, need to add support for that as well.

bittner: Not sure if that's really unimportant. OTOH, we don't handle dates on the server side with analytical, do we? Is any such logic all in the JavaScript code of the services, maybe?  Can we think of any tests the might make sense?

tox.ini
```diff
@@ -10,8 +11,9 @@ deps =
     coverage
     pytest-django
     django22: Django>=2.2,<3.0
-    django30: Django>=3.0,<3.1
     django31: Django>=3.1,<3.2
+    django32: Django>=3.2,<4.0
+    django40: Django>=4.0a1,<4.2
```
bittner, line 16: Shouldn't that be `<4.1`?

bittner (changes requested):

bittner:
BTW, @smithdc1 would it make sense for you to become co-maintainer, maybe?

tox.ini
smithdc1: Yes. 🤦‍♂️

bittner (approved):

smithdc1:
Thanks for merging this. 👍

> BTW, @smithdc1 would it make sense for you to become co-maintainer, maybe? 

I'm interested. Are there notes/docs somewhere on what this means?

bittner:
> I'm interested. Are there notes/docs somewhere on what this means?

That simply means you get more work. :smirk:  Less sleep & more responsibility. :dizzy_face: :1st_place_medal: 

@jezdez Is there anything specific that needs to be done to make @smithdc1 a (co-) maintainer of django-analytical?

bittner:
@smithdc1 Can you please open a [project lead issue](https://github.com/jazzband/help/issues/new/choose) as described [in the Jazzband docs](https://jazzband.co/about/releases#security)?

bittner:
> @smithdc1 Can you please open a [project lead issue](https://github.com/jazzband/help/issues/new/choose) as described [in the Jazzband docs](https://jazzband.co/about/releases#security), asking to become [a project lead](https://jazzband.co/projects/django-analytical) for django-analytical?

@smithdc1 Please. Thank you!

PR #198, opened 2021-10-06
Add Python 3.10 and drop Django 3.1

smithdc1:
Hey, hope all is well. 

This PR adds Python 3.10 supprt and also drops support from Django 3.1 which is now EOL. 

Following review and merge of this do you think we could issue a new release to confirm support of these new versions?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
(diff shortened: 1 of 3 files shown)


smithdc1:
A short update as time has moved on. 3.2.9 and 4.0b1 are both released and support Python 3.10
https://pypi.org/project/Django/3.2.9/
https://pypi.org/project/Django/4.0b1/

bittner (approved):

PR #206, opened 2022-03-06
Consolidate tox.ini and pyproject.toml

bittner:
Just a little cleanup and modernization of `tox.ini` with moving some configuration bits to `pyproject.toml`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,63 +10,6 @@ envlist =
     readme
     docs
 
-[testenv]
-description = Unit tests
-deps =
-    coverage
-    pytest-django
-    django22: Django>=2.2,<3.0
-    django32: Django>=3.2,<4.0
-    django40: Django>=4.0,<4.1
-commands =
-    coverage run -m pytest
-    coverage xml
-
-[testenv:bandit]
-description = PyCQA security linter
-deps = bandit<1.6
-commands = -bandit --ini tox.ini {posargs:-r .}
-
-[testenv:docs]
-description = Build the HTML documentation
-deps = sphinx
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -42,41 +55,25 @@ commands = flake8 {posargs}
...
 
 [testenv:readme]
 description = Ensure README renders on PyPI
-deps = twine
+deps =
+    build
+    twine
 commands =
-    {envpython} setup.py -q sdist bdist_wheel
+    python -m build
```
smithdc1, line 65: Nice 👍

smithdc1 (approved):

smithdc1:
Looks good to me. Looking at this I realise we should likely drop Py3.6 and Django2.2 (it's eol in less than a month).

bittner:
> Looking at this I realise we should likely drop Py3.6 and Django2.2 (it's eol in less than a month).

Good point! Let's be good though to people that love sticking to legacy software (probably 95% of commercial users) and give them time to migrate. Haha. :smile:

bittner:
> Could do with getting a release out so there's a release with all the features of the past couple of years before dropping it?

Excellent point! We only need to update the package metadata (probably on the [version number](https://github.com/jazzband/django-analytical/blob/main/analytical/__init__.py#L7)) and the [change log](https://github.com/jazzband/django-analytical/blob/main/CHANGELOG.rst), and push a version tag. That should do it. – Do you want to go about it?

Not sure about the new version number. Did we have breaking changes, or was it mainly reformatting (Black) and refactoring (UnitTest ➜ pytest)?  ([changes since v3.0.0](https://github.com/jazzband/django-analytical/compare/v3.0.0..main))

PR #207, opened 2022-03-11
Use pathlib instead of os.path (we only support Python 3.6+)

bittner:
pathlib's Path is the more elegant, object-oriented approach for path operations, available since Python 3.6.

We have dropped support for Python < 3.6 for quite a while, so using it for installing the package should be fine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,8 @@
-import os
+"""
+Packaging setup for django-analytical.
+"""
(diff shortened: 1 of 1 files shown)


smithdc1 (approved):

bittner:
Thanks for your approval, @smithdc1!

PR #217, opened 2022-07-14
Add the missing get_identity prefix to google_analytics_gtag, test it

PetrDlouhy:
Add the missing prefix in `get_identity()` in `google_analytics_gtag` and test it.

Change of docs and other tests will come in separate PRs.

PR #219, opened 2022-07-14
Add more information about overriding identity to documentation

PetrDlouhy:


PetrDlouhy:
The docs expects, that #217 is merged.

bittner (changes requested):
Looks good! – I hope you don't mind that I edit your changes so much.

docs/features.rst
```diff
@@ -68,3 +68,23 @@ and is enabled by default.  To disable:
...
+
+or in the template::
+
+    {% with analytical_identity=request.user.uuid|default:None %}
+        {% analytical_head_top %}
+    {% endwith %}
+
+If you want to change the identity only for specific provider, use the
+`*_identity` context variable, where `*` is prefix specified on the provider
+page.
```
bittner, line 90:
```suggestion
If you want to change the identity only for specific provider use the
``*_identity`` context variable, where ``*``-prefix is the module name
of the specific provider.
```

bittner:
I believe, we should also mention [context processors](https://docs.djangoproject.com/en/4.0/ref/templates/api/#writing-your-own-context-processors) as an option.

## Example

```python
# FILE: myproject/context_processors.py
from django.conf import settings

def get_identity(request):
    return {
        'analytical_identity': 'some-value-here',
    }
```

With your project's `settings.py` like this:

```python
# FILE: myproject/settings.py
TEMPLATES = [
    {
        'OPTIONS': {
            'context_processors': [
                'myproject.context_processors.get_identity',
            ],
        },
    },
]
```

That allows you as a developer to leave your view code untouched _and_ make sure that the variable is injected for _all_ templates.

PetrDlouhy:
Hope all improvements are now in the PR.

docs/features.rst
```diff
@@ -68,3 +68,54 @@ and is enabled by default.  To disable:
     Alternatively, add one of the variables to the context yourself
     when you render the template.
 
+
+
```
bittner, line 72:
```suggestion

```

Two empty lines should be enough.

docs/features.rst
```diff
@@ -68,3 +68,54 @@ and is enabled by default.  To disable:
...
+or in the template:
+
+.. code-block:: django
+
+    {% with analytical_identity=request.user.uuid|default:None %}
+        {% analytical_head_top %}
+    {% endwith %}
+
+or in context proessor:
```
bittner, line 93:
```suggestion
or by implementing a context processor:
```

docs/features.rst
```diff
@@ -68,3 +68,54 @@ and is enabled by default.  To disable:
...
+                ],
+            },
+        },
+    ]
+
+That allows you as a developer to leave your view code untouched and
+make sure that the variable is injected for all templates.
+
+If you want to change the identity only for specific provider use the
+``*_identity`` context variable, where ``*``-prefix is the module name
```
bittner, line 120:
```suggestion
``*_identity`` context variable, where the ``*`` prefix is the module name
```

bittner (changes requested):
Great, thanks! Just another few corrections, then we're ready to merge.

bittner (approved):

bittner:
Shoot! I should have squashed the commits... :worried:

bittner:
Thanks for your contribution! :100:

PR #227, opened 2023-05-02
Update CI setup.

jezdez:
- remove support for Django 2.2
- remove support for Django 4.0
- add support for Django 4.2
- add support for Python 3.11 on supported Django versions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,24 +10,7 @@ on:
 
 jobs:
   python-django:
-    runs-on: ubuntu-latest
-    strategy:
-      max-parallel: 5
-      matrix:
-        python-version:
-        - '3.6'
-        - '3.7'
-        - '3.8'
-        - '3.9'
-        - '3.10'
-        django-version:
-        - '2.2'
-        - '3.2'
-        - '4.0'
-        exclude:
-        - { django-version: '2.2', python-version: '3.10' }
-        - { django-version: '4.0', python-version: '3.6' }
-        - { django-version: '4.0', python-version: '3.7' }
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -18,23 +18,10 @@ jobs:
     - uses: actions/checkout@v2
```
bittner, line 18:
```suggestion
    - uses: actions/checkout@v3
```

.github/workflows/test.yml
```diff
@@ -18,23 +18,10 @@ jobs:
...
-      with:
-        path: ${{ steps.pip-cache.outputs.dir }}
-        key:
-          ${{ matrix.python-version }}-v1-${{ hashFiles('**/setup.py') }}
-        restore-keys: |
-          ${{ matrix.python-version }}-v1-
+        cache: 'pip'  # caching pip dependencies
 
     - name: Install dependencies
       run: |
```
bittner, line 27: We may also skip the Pip version check, relying on GitHub to ship a fairly recent, continuously updated version. E.g.

```bash
python -m pip install tox-gh-actions wheel --disable-pip-version-check
```

bittner (changes requested):

.github/workflows/test.yml
```diff
@@ -9,55 +9,46 @@ on:
     - main
 
 jobs:
-  python-django:
-    runs-on: ubuntu-latest
+  test:
+    runs-on: ubuntu-20.04
```
bittner, line 13:
```suggestion
    runs-on: ubuntu-latest
```

We should simply drop Python 3.6 support. GitHub's VMs for GHA have stopped shipping unsupported Python versions. That makes things difficult.

bittner (changes requested):
This PR needs to be finished and merged.

bittner:
@jezdez Do you want to push this forward? A full overhaul would be overdue in any case.

PR #229, opened 2023-09-22
Use newer features of PyClean

bittner:
PyClean now has a `--debris` option that allow us to simplify the related Tox configuration section.


bittner:
The tests for Python 3.6 fail, because the VMs for GHA don't ship unsupported Python versions anymore.

bittner:
Implemented via PR #233.

PR #230, opened 2024-12-18
Port to Sphinx 8.0

cjwatson:
The old `intersphinx_mapping` format has been removed; it must now map identifiers to (target, inventory) tuples.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,9 +28,11 @@ add_function_parentheses = True
 pygments_style = 'sphinx'
 
 intersphinx_mapping = {
-    'http://docs.python.org/2.7': None,
-    'http://docs.djangoproject.com/en/1.9':
-        'http://docs.djangoproject.com/en/1.9/_objects/',
+    'python': ('https://docs.python.org/3', None),
+    'django': (
+        'https://docs.djangoproject.com/en/dev',
+        'https://docs.djangoproject.com/en/dev/_objects/',
+    ),
 }
 
 


docs/conf.py
```diff
@@ -28,9 +28,11 @@
 pygments_style = 'sphinx'
 
 intersphinx_mapping = {
-    'http://docs.python.org/2.7': None,
-    'http://docs.djangoproject.com/en/1.9':
+    'python': ('http://docs.python.org/3', None),
+    'django': (
+        'http://docs.djangoproject.com/en/1.9',
         'http://docs.djangoproject.com/en/1.9/_objects/',
```
bittner, line 34: Does this have to be an outdated version of Django?

docs/conf.py
```diff
@@ -28,9 +28,11 @@
 pygments_style = 'sphinx'
 
 intersphinx_mapping = {
-    'http://docs.python.org/2.7': None,
-    'http://docs.djangoproject.com/en/1.9':
+    'python': ('http://docs.python.org/3', None),
```
bittner, line 31:
```suggestion
    'python': ('http://docs.python.org/', None),
```

Maybe?

docs/conf.py
cjwatson: Not particularly.  Fixed.

docs/conf.py
cjwatson: I changed it as you suggest, but note that it results in this message while building the docs:

```
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
```

... so let me know if you want me to change it back in light of that.

bittner: In that case please revert it, sure. Didn't know.

But updating the protocol to https in all URLs would make sense, wouldn't it?

cjwatson: OK, done on both counts.

bittner:
Unfortunately, the pipeline jobs fail all over the place. Not your fault, but I'm not sure if the CI setup shouldn't better be overhauled first. :thinking: :thought_balloon: 

See also #227. for example.

bittner:
I opened PR #233 to unblock all PRs that are currently blocked by our outdated CI setup. Doing so, I unwittingly integrated your suggested changes. Thank you for your effort! :1st_place_medal: 

Closing this PR as now obsolete/duplicate.

PR #100, opened 2016-12-12
added config options for Woopra

retosteffen:
config option as per
https://www.woopra.com/docs/setup/javascript-tracking/
added as WOOPRA_XXXX (uppercase of the config name)

PR #239, opened 2025-07-09
Add more config options for Woopra

bittner:
Closes #100 (abandoned)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Unreleased
   license metadata field (Peter Bittner)
 * Remove AUTHORS file to avoid confusion; this is now metadata maintained
   in pyproject.toml (Peter Bittner)
+* Add more configuration options for Woopra (Peter Bittner)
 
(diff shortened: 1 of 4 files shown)


bittner:
Using `contextlib.suppress` is the pythonic way ([according to Raymond Hettinger](https://www.youtube.com/watch?v=OSGv2VnC0go&t=2603s)). It's basically `try: ... except: pass`. In comparison to a loop with `if` it avoids a potential race condition, so I'd prefer to keep it.

> Failing on invalid values is probably fine? `ValueError` isn't handled after all. That said, are you sure you need to coerce the values?

True, checking for validity would be more user-friendly. It's a lazy solution. The type casting forces some errors, causing a `ValueError`, that's the only idea behind it. So, fair enough, I'll add type checking into the code.

bittner:
Thank you, Mathias, for the review! :bow: 

I don't know Woopra either, I just wanted to finish off an abandoned PR. I think we did a good job on this.

PR #240, opened 2025-07-09
Release v3.2.0

bittner:
This is to prepare a new release on PyPI.

If I remember correctly, I need to merge this PR and only then push a version tag on the resulting commit on the `main` branch.

In accordance with the current Git tags in this repo, this will be `v3.2.0` (note the leading `v`). Does this look fine for you all?

## References

- [Jazzband Releases](https://jazzband.co/about/releases#continuous-integration)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,5 @@
-Unreleased
-----------
+Version 3.2.0
+-------------
 * Remove deprecated Piwik integration. Use Matomo instead! (Peter Bittner)
 * Migrate packaging from setup.py to pyproject.toml with Ruff for linting
   and formatting (Peter Bittner)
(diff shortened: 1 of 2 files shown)


matthiask (approved):
Let's trust the people who are putting the work into this package.

bittner:
This is bad! The problem stems from the [release.yml](https://github.com/jazzband/django-analytical/blob/main/.github/workflows/release.yml#L40) GHA configuration. We need to fix this before the release can work. :confused: 

image

bittner:
Interestingly, the release worked despite the "error message" in the [build job](https://github.com/jazzband/django-analytical/actions/runs/16207031141). The green tick was correct as an indicator.

I had to release the package on the [Jazzband project page](https://jazzband.co/projects/django-analytical), afterwards. The release is now public [on PyPI](https://pypi.org/project/django-analytical/3.2.0/).

